Issue #18, opened 2014-12-21
Feature request: remove duplication in package_read author and maintainer email


florianm:
Follow [package/snippets/additional_info](https://github.com/ckan/ckan/blob/master/ckan/templates/package/snippets/additional_info.html) and roll author/maintainer name/email into one field [here](https://github.com/open-data/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/package/snippets/additional_info.html), ideally avoiding the use of `c` in the snippet as per https://github.com/ckan/ckan/issues/2132

PR #22, opened 2014-12-27
#18 simplify author and maintainer name and email into one field each

florianm:
As per #18 - template code from [ckan's](https://github.com/ckan/ckan/blob/master/ckan/templates/package/snippets/additional_info.html#L19) additional fields works nicely at the cost of introducing usage of `pkg_dict` and breaking the sequential handling of fields.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+omit = */tests/*
+source = ckanext
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,2 @@
 [run]
-omit = */tests/*
-source = ckanext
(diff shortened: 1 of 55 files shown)


florianm:
Actually there's an even simpler way. "author" gets hidden with `'display_field': false`, and "author_email" gets rendered as name (if no email given), email (if no name given) or name as email link (if both given) via `"display_email_name_field": "author"`. No template changes necessary, so that scales up nicely. 

Also I'll close this PR and submit a cleaner PR shortly. Note to self, no merges before coffee.

florianm:
Replaced with a cleaner PR #23

PR #23, opened 2015-01-06
#18 combine name and email display

florianm:
As discussed in (messy and closed) PR #22 this PR combines the display of name/email combo fields by hiding the name field, and letting the email field render the email address (if given) with the specified name field (if given).

### DIFF
--- a/ckanext/scheming/ckan_dataset.json
+++ b/ckanext/scheming/ckan_dataset.json
@@ -55,7 +55,8 @@
       "field_name": "author",
       "label": "Author",
       "form_placeholder": "Joe Bloggs",
-      "display_property": "dc:creator"
+      "display_property": "dc:creator",
+      "display_field": false
     },
     {
       "field_name": "author_email",
@@ -69,7 +70,8 @@
       "field_name": "maintainer",
       "label": "Maintainer",
       "form_placeholder": "Joe Bloggs",
-      "display_property": "dc:contributor"
+      "display_property": "dc:contributor",
+      "display_field": false
     },
     {
       "field_name": "maintainer_email",


wardi:
we still need the thing that checks for `'display_field': False`, right?  Alternatively we could use `'display_snippet': null`.

florianm:
Strangely, `display_field: false` just works - although I can't find where that happens, at least there's nothing acting upon `display_field` or manipulating exclude fields in ckan or scheming as far as I can grep. Any ideas?

wardi:
Hmm. doesn't seem to hide the fields when I try it.

florianm:
Found why it works on my end - I'm still using my much too clever [additional info display snippet](https://github.com/florianm/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/package/snippets/additional_info.html) with special treatment for author and maintainer fields.

wardi:
@florianm ok, should we close this too, or will you add the display_snippet is None check to this PR?

florianm:
@wardi sorry I got sidetracked - I will add the check to this PR today.

florianm:
@wardi added a property "hide_field", formatted additional_info.html for readability and added check for hide_field there, added email example to camel_photos.json.

Up for discussion:

* The author-hiding property is named `hide_field` (instead of `display_field`) so the template can check for its existence and whether to hide the field in one operation.
* I've added values for the email example to tests which would otherwise fail. I can create a dataset without author and email fields through the GUI, but the tests will fail on blank author/email.

florianm:
@wardi I need to double-check whether dataset notes are hidden inadvertently, sorry that's one thing I might have clobbered last minute!

florianm:
@wardi there's travis errors with the solr server that I'm sure I didn't touch. Any ideas whether that's a problem to do with my changed templates?
Also, please allow me a few days of thinking (my debugging appears to slow down when nappie changing instead of sleeping) to figure out why dataset notes are not showing on my end.

Issue #24, opened 2015-01-09
Format date display and form field to hide time


florianm:
CKAN saves datetimes, but the jquery-ui datepicker shows only dates and will be blank if set to a datetime. This causes a dataset with date fields to lose their value when updating.

The new date preset should format CKAN's datetime to dates only, so the datepicker won't lost the value, and the date field is, as the name says, just a date field.

wardi:
closed via #25

PR #25, opened 2015-01-09
format existing datetime to date only in form and display snippets

florianm:
Addressing #24 this hides the time component of CKAN's datetime fields in ckanext-scheming's date preset.

### DIFF
--- a/ckanext/scheming/presets.json
+++ b/ckanext/scheming/presets.json
@@ -81,6 +81,7 @@
       "preset_name": "date",
       "values": {
         "form_snippet": "date.html",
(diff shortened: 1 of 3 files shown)


wardi:
@florianm Given https://github.com/ckan/ckan/issues/2228 you might want to rework this to not call render_datetime and just build the YYYY-MM-DD date directly.

florianm:
@wardi thanks for the heads-up. Hope I'll get to it later this week!

florianm:
should have listened to @wardi - dates before 1900 _do_ blow up.

PR #27, opened 2015-01-19
Translated  package state in additional_info.html

andreivicentiuhincu:
The state is now translated.


wardi:
I'll take this, but you should really also submit the same change upstream: https://github.com/ckan/ckan/blob/master/ckan/templates/package/snippets/additional_info.html#L57

PR #29, opened 2015-01-20
Group & Org templates

deniszgonjanin:
Work in progress to make Group and Org schemas editable through the web templates


ckanext/scheming/plugins.py
```diff
@@ -118,20 +118,16 @@ class _GroupOrganizationMixin(object):
     def group_types(self):
         return list(self._schemas)
 
-    def setup_template_variables(self, context, data_dict):
-        group_type = context.get('group_type')
+    def setup_template_variables(self, context, data_dict, group_type):
```
deniszgonjanin: @wardi I had to get CKAN to explicitly pass the group type to the plugin here, since it wasn't being passed into the context before (`ckan.controllers.group.py:59`). The way it is now works for me, but is not backwards-compatible with other IGroupForm plugins, so another fix in core CKAN would be needed, like explicitly forcing the group_type into the context object.

PR #31, opened 2015-01-21
Change of scheming_required

benjaminlaot:
See : https://github.com/open-data/ckanext-scheming/issues/30

Change the validator not_missing by the  not_empty one


wardi:
Good catch. I'll merge but a test would be appreciated too.

Issue #30, opened 2015-01-21
scheming_required apply the not_missing validator instead of the not_empty one


benjaminlaot:
When I set "required" : "true" and as my first validator : "scheming_required", nothing happened because  the validator not_missing is present here instead of not_empty :

@scheming_validator
def scheming_required(field):
    """
    not_empty if field['required'] else ignore_missing
    """
    if field.get('required'):
        return not_missing
    return ignore_missing

PR #32, opened 2015-01-30
#30 Add Test for scheming_required, wrong validator is passed and fix _field_validators

benjaminlaot:
Fix of _field_validators that used "not_missing" too instead of "not_empty"

Add a test for scheming required.


benjaminlaot:
I forgot to define scheming_required.
NameError: global name 'scheming_required' is not defined

PR #38, opened 2015-02-27
fix gettext not encoding as utf-8

deniszgonjanin:


wardi:
this is the opposite: gettext is returning a binary string in utf-8 encoding instead of a unicode string. This decodes that binary string to unicode for safe use in templates.

We should have a short test for this to make sure it stays fixed (it should be about 3 lines, this function lets you pass in your own gettext function)

PR #59, opened 2015-07-23
These changes allow the date preset to handle dates < 1900

marcfor:
Should we use data.get() in the display snippet as well?


wardi:
@marcfor I think the template that renders the display snippets skips them if the value is missing

PR #60, opened 2015-07-28
Fix exception in scheming_language_text.

TkTech:
If the dict passed to scheming_language_text was empty,
it would crash when it tries getting the first result
from sorted().

Short-circuit an empty "text" value. Also uses exceptions
instead of get() chains (performs better when not using a
default on get(), and in the case where you *usually* expect
there to be a value for that key.)


wardi:
dang. missed my chance to ask for a regression test. Next time I won't be so lenient!

PR #62, opened 2015-07-29
Prefix config options with 'ckanext.'

amercader:
To avoid clashes with other extensions or external libraries and to keep
things tidy.

The old `scheming.*` options are still fully supported.

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -83,13 +83,21 @@ class _SchemingMixin(object):
     def _load_presets(self, config):
         if _SchemingMixin._presets is not None:
             return
-        presets = config.get('scheming.presets', DEFAULT_PRESETS).split()
+        presets = config.get('ckanext.scheming.presets', DEFAULT_PRESETS).split()
         _SchemingMixin._presets = {}
         for f in reversed(presets):
             for p in _load_schema(f)['presets']:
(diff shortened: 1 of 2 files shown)


ckanext/scheming/plugins.py
```diff
@@ -83,13 +83,21 @@ def _add_template_directory(self, config):
...
                 _SchemingMixin._presets[p['preset_name']] = p['values']
 
     def update_config(self, config):
+
+        # Check for old 'scheming.*' config options
+        new_options = {}
+        for key, value in config.iteritems():
+            if key.startswith('scheming.'):
+                new_options['ckanext.' + key] = value
+        config.update(new_options)
```
wardi: @amercader This is going to break all the config options you're worried about us conflicting with!

amercader: @wardi ok ok, I can add another check to see if there's an existing `'ckanext.' + key` before  overwriting it :)

The whole PR is no big deal, I'm happy to drop it if you want, I just like extensions config options to start with `ckanext.`

wardi: it's so much longer, and my helpers are all scheming_... and my actions are all scheming_... I'd rather keep it the way it is unless ckanext.<foo> really is the accepted standard

PR #68, opened 2015-07-30
Added new helpers to retrieve presets.

TkTech:
Added scheming_get_preset(s) to retrieve any defined presets.

Also removed unused imports, trambled imports, and trambled variable
from plugins.py.

### DIFF
--- a/ckanext/scheming/helpers.py
+++ b/ckanext/scheming/helpers.py
@@ -74,6 +74,30 @@ def scheming_dataset_schemas(expanded=True):
         return p.instance._schemas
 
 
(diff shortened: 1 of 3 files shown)


wardi:
@TkTech Would you do the pep8 stuff as a separate commit? Mixing pep8 and real changes is going to make git blame output confusing

wardi:
Great, so a couple little tests and I'll merge this

thriuin:
@wardi Hi Ian, I'm going to go ahead and merge this in because we could use it this week.

PR #76, opened 2015-08-24
Fix a template using cached values instead of post-processed values.

TkTech:
This issue is also in CKAN core. Since scheming replicates the template, it inherits this bug.


wardi:
:+1: will merge when the tests pass

wardi:
see also https://github.com/ckan/ckan/pull/2595 upstream fix

PR #81, opened 2015-09-03
Fix `scheming` command to allow dataset_fields key.

brew:
Group and Org schemas use `fields` as the key to the array within the
schema json file. Dataset schemas use `dataset_fields`. This commit
stops the `scheming` command from breaking when printing dataset fields.

### DIFF
--- a/ckanext/scheming/commands.py
+++ b/ckanext/scheming/commands.py
@@ -40,14 +40,28 @@ class SchemingCommand(CkanCommand):
             print self.__doc__
 
     def _show(self):
-        for n, s in (
-                ("Dataset", scheming_dataset_schemas()),
-                ("Group", scheming_group_schemas()),
-                ("Organization", scheming_organization_schemas()),
-                ):
+        dataset_schema = scheming_dataset_schemas()
+
+        resource_schema = None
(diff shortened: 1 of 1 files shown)


brew:
Go on then ;)

wardi:
If I'm reading this correctly this will separate the resource fields from their corresponding dataset fields when multiple dataset schemas are enabled. That might be confusing. Can you instead list the resource fields right after the dataset fields for a given schema?

brew:
It treats the resource schema like a separate schema. The output is after datasets, like this:

```
Dataset schemas:
 * "dataset"
   - "fluent_title"
   - "name"
   - ...

Resource schemas:
 * "resource"
   - "url"
   - "fluent_name"
   - "fluent_description"
   - "format"

Group schemas:
 * "group"
   - "title"
   - "name"
   - "notes"
   - "url"
   - "bookface"

Organization schemas:
    plugin not loaded or schema not specified
```

But, yes, we could have it so resources appear with "dataset" like this:

```
Dataset schemas:
 * "dataset"
   - "fluent_title"
   - "name"
   - ...

 * "resource"
   - "url"
   - "fluent_name"
   - "fluent_description"
   - "format"
...
```

PR #82, opened 2015-09-04
Move test.ini and add README notes for tests

### LINKED ISSUE #63: Document / make easier to run the tests locally
The `ini` files are located on the tests folder, which is fine but not where generally extensions have them, so worth mentioning on the README.
The `use = config:../../../links/ckan/test-core.ini` has a links folder not present on a default set up, I guess this is set for Travis?
There are two ini files, what's `test_subclass.ini` for?

brew:
This moves test.ini to the standard location, assuming the extension is installed beside ckan. Fixes #63.


wardi:
looks great

PR #83, opened 2015-09-09
Group templates

brew:
This continues @deniszgonjanin's work on Group templates for forms and display (#29).

The main thing here is that convert_from_extras wasn't working for groups with the standard converter, so I've added one to scheming tailored to groups.

Also, the test for group_form for new groups should pass once ckan/ckan#2617 is merged [1]

[1] https://github.com/brew/ckanext-scheming/blob/59c5197f850c94b14b4b1210a192dda1bc350bb8/ckanext/scheming/tests/test_form.py#L77


deniszgonjanin:
awesome. I'm going to try this out

brew:
Warning that Organizations is still bust if using `scheming_organizations` plugin, with a rather cryptic `Exception: menu item 'organization_activity' need parameter 'offset'` error when viewing an organization. The new form displays correctly though.

deniszgonjanin:
Works great for groups

deniszgonjanin:
Could be merged as is @wardi, polishing up the org support could be done in another PR

PR #88, opened 2015-09-29
Added a datetime preset for using HTML5 date and time elements

LaurentGoderre:
TODO: Separate the validation logic to validate the date and time separately

### DIFF
--- a/ckanext/scheming/templates/scheming/display_snippets/datetime.html
+++ b/ckanext/scheming/templates/scheming/display_snippets/datetime.html
@@ -1,0 +1,1 @@
+{% extends 'scheming/display_snippets/text.html' %}
--- a/ckanext/scheming/templates/scheming/form_snippets/datetime.html
+++ b/ckanext/scheming/templates/scheming/form_snippets/datetime.html
@@ -1,0 +1,42 @@
+{% import 'macros/form.html' as form %}
+{% set date = data.get(field.field_name + '_date') %}
+{% set time = data.get(field.field_name + '_time') %}
+
+{% if not date is string and not time is string %}
+    {% set date = data.get(field.field_name) %}
+
+    {% if date %}
+        {% set parts = data.get(field.field_name).split() %}
+        {% set date = parts[0] %}
+        {% set time = parts[1] %}
+    {% endif %}
+{% endif %}
+{% call form.input(
+    field.field_name + '_date',
+    id='field-' + field.field_name +  '-date',
+    label= h.scheming_language_text(field.label),
+    type='date',
+    value=date,
+    error=errors[field.field_name + '_date'],
+    classes=['control-medium'],
+    attrs=field.form_attrs if 'form_attrs' in field else {},
+    is_required=h.scheming_field_required(field)
+    )
+%}
+    {%- snippet 'scheming/form_snippets/help_text.html', field=field -%}
+{% endcall %}
+
+{% call form.input(
+    field.field_name + '_time',
+    id='field-' + field.field_name + '-time',
+    label= h.scheming_language_text(field.get('label_time', 'Time')),
+    type='time',
+    value=time,
+    error=errors[field.field_name + '_time'],
+    classes=['control-medium'],
+    attrs=field.form_attrs if 'form_attrs' in field else {},
+    is_required=h.scheming_field_required(field)
+    )
+%}
+    {%- snippet 'scheming/form_snippets/help_text.html', field=field -%}
+{% endcall %}
--- a/ckanext/scheming/tests/test_validation.py
+++ b/ckanext/scheming/tests/test_validation.py
@@ -133,6 +133,158 @@ class TestDates(object):
                 'camels_in_photo': 2,
                 'date': '2015-01-01'}])
 
+class TestDateTimes(object):
+    def test_datetime_field_rejects_non_isodates(self):
+        lc = LocalCKAN()
+        try:
+            lc.action.package_create(
+                type='camel-photos',
+                name='fred_datetime1',
+                a_relevant_datetime='31/11/2014',
+            )
+        except ValidationError as e:
+            assert_equals(e.error_dict['a_relevant_datetime'],
+                ['Date format incorrect']
+        )
+        else:
+            raise AssertionError('ValidationError not raised')
+
+        try:
+            lc.action.package_create(
+                type='camel-photos',
+                name='fred_datetime2',
+                a_relevant_datetime='31/11/abcd',
+            )
+        except ValidationError as e:
+            assert_equals(e.error_dict['a_relevant_datetime'],
+                ['Date format incorrect']
+        )
+        else:
+            raise AssertionError('ValidationError not raised')
+
+        try:
+            lc.action.package_create(
+                type='camel-photos',
+                name='fred_datetime3',
+                a_relevant_datetime='this-is-not-a-date',
+            )
+        except ValidationError as e:
+            assert_equals(e.error_dict['a_relevant_datetime'],
+                ['Date format incorrect']
+        )
+        else:
+            raise AssertionError('ValidationError not raised')
+
+        try:
+            lc.action.package_create(
+                type='camel-photos',
+                name='fred_datetime4',
+                a_relevant_datetime='2014-11-15Tabcd',
+            )
+        except ValidationError as e:
+            assert_equals(e.error_dict['a_relevant_datetime'],
+                ['Date format incorrect']
+        )
+        else:
+            raise AssertionError('ValidationError not raised')
+
+        try:
+            lc.action.package_create(
+                type='camel-photos',
+                name='fred_datetime5',
+                a_relevant_datetime='2014-11-15T12:00:ab',
+            )
+        except ValidationError as e:
+            assert_equals(e.error_dict['a_relevant_datetime'],
+                ['Date format incorrect']
+        )
+        else:
+            raise AssertionError('ValidationError not raised')
+
+    def test_date_field_valid_date_str(self):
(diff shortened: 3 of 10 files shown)


ckanext/scheming/validation.py
```diff
@@ -1,9 +1,14 @@
 import json
+import datetime
+import ckan.lib.helpers as h
+import ckan.lib.navl.dictization_functions as df
```
wardi: the coding standards for ckan extensions says not to import directly from ckan, use the toolkit modules instead

ckanext/scheming/validation.py
```diff
@@ -83,6 +88,27 @@ def validator(key, data, errors, context):
     return validator
 
 
+@scheming_validator
+def scheming_isodatetime(value, context):
+    def validator(key, data, errors, context):
+        value = data[key]
+        if type(value) == list:
+            value = data[key] = value[0] + ' ' + value[1]
```
wardi: This will result in a server error if the list has fewer than two elements or if the types in the list aren't unicode or str. You can wrap it in a try/except to catch those cases. Also, `isinstance(value, list)` is a little better than using `type(value) == list`

ckanext/scheming/validation.py
```diff
@@ -83,6 +88,27 @@ def validator(key, data, errors, context):
     return validator
 
 
+@scheming_validator
+def scheming_isodatetime(value, context):
```
wardi: the parameters to scheming validators are the whole schema and the field schema, not value and context. It seems like you're not using these values at all, so don't bother with `@scheming_validator` at all

ckanext/scheming/validation.py
```diff
@@ -83,6 +88,27 @@ def validator(key, data, errors, context):
...
+    def validator(key, data, errors, context):
```
wardi: You could write this as a simple one or two parameter validator instead. Those validators can return the value they've converted. This validator won't do anything because four-parameter validators need to set their value back in data[key] to see any change. http://docs.ckan.org/en/latest/extensions/adding-custom-fields.html?highlight=custom%20validator#custom-validators

ckanext/scheming/templates/scheming/form_snippets/datetime.html
```diff
@@ -0,0 +1,42 @@
...
+{% set time = '' %}
+{% set dateTime = data.get(field.field_name, '') %}
+
+{% if dateTime is string %}
+    {% set parts = dateTime.split() %}
+    {% set date = parts[0] %}
+    {% set time = parts[1] %}
+{% elif dateTime is sequence %}
+    {% set date = dateTime[0] %}
+    {% set time = dateTime[1] %}
```
wardi: this if/elif is unnecessary. The form snippet will always receive the converted string value (from the output validator), not a list.

ckanext/scheming/templates/scheming/form_snippets/datetime.html
```diff
@@ -0,0 +1,42 @@
...
+    {%- snippet 'scheming/form_snippets/help_text.html', field=field -%}
+{% endcall %}
+
+{% call form.input(
+    field.field_name,
+    id='field-' + field.field_name + '-time',
+    label= h.scheming_language_text(field.get('label_time', 'Time')),
+    type='time',
+    value=time,
+    error=errors[field.field_name],
```
wardi: The errors should appear on one field or the other, not both. This will show the same error twice if there is an error on the field.

LaurentGoderre: Actually, it's very bad if the error is on only one field because of the error is in the time field and that field is not identified as having an error, the user will have a hard time figuring out that the issue is with a field not highlighted. Ideally we should narrow the error to one of the other field but I haven't quite figured out how to do that yet.

wardi: If you give the fields different names you can use the 4-argument version of the validator to associate errors with the separate fields. It will only work with the form, but I think that's what you want

wardi:
Would you add this new type to the readme?

A functional test would be nice too so that we know this code stays working.

LaurentGoderre:
Ok! On it!

ckanext/scheming/tests/test_validation.py
```diff
@@ -80,6 +80,35 @@ def test_date_field_in_resource(self):
...
 
+class TestDateTimes(object):
+    def test_datetime_field_rejects_non_isodates(self):
+        lc = LocalCKAN()
+        assert_raises(ValidationError, lc.action.package_create,
+            type='camel-photos',
+            name='fred',
+            a_relevant_datetime='31/11/2014',
+        )
+        assert_raises(ValidationError, lc.action.package_create,
```
wardi: these should be in separate tests (same goes for the date tests above) and how about at least one test of a valid datetime?

ckanext/scheming/templates/scheming/display_snippets/datetime.html
```diff
@@ -0,0 +1 @@
+{{ data[field.field_name].split()[0]+'T'+data[field.field_name].split()[1] }}
```
wardi: you like the "T" in date time displays? :fearful:

ckanext/scheming/validation.py
```diff
@@ -82,6 +85,59 @@ def validator(key, data, errors, context):
...
+            except (TypeError, ValueError), e:
+                raise Invalid(date_error)
+        else:
+            extras = data.get(('__extras',))
+            if not extras or key[0] + '_date' not in extras:
+                if field.get('required'):
+                    not_empty(key, data, errors, context)
+            else:
+                for input_suffix in ['date', 'time']:
+                    input = key[0] + '_' + input_suffix
```
wardi: `input` is a builtin, if you use it as a variable name @TkTech's editor won't let him edit this file :smirk:

ckanext/scheming/validation.py
```diff
@@ -82,6 +85,59 @@ def validator(key, data, errors, context):
...
+        if not value is missing:
+            try:
+                 date = h.date_str_to_datetime(value)
+            except (TypeError, ValueError), e:
+                raise Invalid(date_error)
+        else:
+            extras = data.get(('__extras',))
+            if not extras or key[0] + '_date' not in extras:
+                if field.get('required'):
+                    not_empty(key, data, errors, context)
```
wardi: is there a `return` missing here?

ckanext/scheming/validation.py
```diff
@@ -82,6 +85,59 @@ def validator(key, data, errors, context):
...
+                    elif input_suffix == 'time' and not value == '':
+                        try:
+                            value_full += ' ' + value
+                            date = h.date_str_to_datetime(value_full)
+                        except (TypeError, ValueError), e:
+                            errors[new_key].append(time_error)
+                            #raise Invalid(time_error)
+
+        data[key] = date
+        return date
```
wardi: I'm pretty sure the return value from 4-argument validators is ignored

ckanext/scheming/validation.py
```diff
@@ -82,6 +85,59 @@ def validator(key, data, errors, context):
...
+                 date = h.date_str_to_datetime(value)
+            except (TypeError, ValueError), e:
+                raise Invalid(date_error)
+        else:
+            extras = data.get(('__extras',))
+            if not extras or key[0] + '_date' not in extras:
+                if field.get('required'):
+                    not_empty(key, data, errors, context)
+            else:
+                for input_suffix in ['date', 'time']:
```
wardi: if you've got a block of code you want to reuse you can create a nested function to call instead of a for loop with constant values. That will also save you the awkward `if input_suffix` checks below. Your nested functions have access to the locals in this function too.

ckanext/scheming/templates/scheming/display_snippets/datetime.html
LaurentGoderre: Not really no, but it didn't occur it was our client's preference. I can make it generic and override it in our project

ckanext/scheming/validation.py
LaurentGoderre: No, the return is at the end

ckanext/scheming/validation.py
LaurentGoderre: Yeah, it seem so. I was just following the other validators example

ckanext/scheming/tests/test_validation.py
LaurentGoderre: Well, they are all invalid dates so it can make sense as all being togheter

ckanext/scheming/templates/scheming/display_snippets/datetime.html
LaurentGoderre: What do you think it should be? The same without the `T`?

ckanext/scheming/validation.py
LaurentGoderre: I don't see how converting to a input function would remove the need for the `if`....

ckanext/scheming/templates/scheming/display_snippets/datetime.html
```diff
@@ -0,0 +1,2 @@
+{% set url_parts = data[field.field_name].split() %}
+{{ url_parts[0] }} {{ url_parts[1] }}
```
wardi: `{{ data[field.field_name] }}` is the same as these two lines, right? which is also the same as just using the existing "text.html" display_snippet

LaurentGoderre: How about replacing it with `{% extends 'scheming/display_snippets/text.html' %}` which would allow us to overwrite it if we need?

PR #89, opened 2015-10-01
Simple template updates to support 'default' values in schemas.

TkTech:


wardi:
The c.action test doesn't work for new resources, right?

wardi:
@TkTech could you add some defaults to the camel-photos schema and a couple simple tests to https://github.com/open-data/ckanext-scheming/blob/master/ckanext/scheming/tests/test_form.py#L34 that show the defaults appearing in datasets and resources? (if the resources aren't working use SkipTest for now)

wardi:
@TkTech so there's no way to make this approach work with resources without a core ckan change? I might just switch this to testing for the presence of the fields in the data passed, unless you already know that doesn't work

TkTech:
@wardi Yes, we can make it work with resources, I just haven't had time to do it yet.

wardi:
ok, thanks

wardi:
Just needs a description of the feature in the README and small tests in https://github.com/open-data/ckanext-scheming/blob/master/ckanext/scheming/tests/test_form.py that defaults appear and don't overwrite existing values in the dataset and resource forms

TkTech:
Better ways to do this I think and we no longer need it for NDM.

PR #96, opened 2015-10-21
Fix validation tests

LaurentGoderre:


wardi:
@LaurentGoderre try `from __future__ import with_statement` at the top of that module to fix the 2.6 error

LaurentGoderre:
@wardi: it would be nice but no, @TkTech figured out that in 2.6 `assert_raises` doesn't act as a context manager. We figured out an alternate method

wardi:
but we're using nose's `assert_raises`... Why nose? Why do you leak python implementation details?

TkTech:
@wardi https://github.com/nose-devs/nose/blob/master/nose/tools/trivial.py#L33

It's literally just renaming the methods in unittest. Sadly, the documentation for assertRaises is such:

> Changed in version 2.7: Added the ability to use assertRaises() as a context manager.

PR #98, opened 2015-10-30
Support Schema File Automatic Reloading (& PEP8 plugins.py)

TkTech:


ckanext/scheming/plugins.py
```diff
@@ -340,6 +350,7 @@ def _load_schema_module_path(url):
         return
     p = os.path.join(os.path.dirname(inspect.getfile(m)), file_name)
     if os.path.exists(p):
+        watch_file(p)
```
wardi: We don't need anything conditional here to see if we're actually running under paster serve?

TkTech: https://bitbucket.org/ianb/paste/src/a159cd816e016984398a9f4332777fee6d0dce47/paste/reloader.py?at=default&fileviewer=file-view-default#reloader.py-128

Yup. Reloader is a singleton that will always exist.

wardi:
love it

PR #99, opened 2015-10-30
Added LoaderError exception, and a helpful error message if…

TkTech:
… attempting to load a YAML file without PyYAML.


wardi:
I did https://github.com/wardi/ckanext-scheming/commit/93af96f instead

PR #102, opened 2015-11-13
Fixed scheming_language_text failing when being called from ckanapi

LaurentGoderre:


wardi:
I don't get it.

What's this a fix for? Could you write a test that causes the error you're seeing?

LaurentGoderre:
Well basically, since the fluent core output validator, when creating datasets throught the api, it's trying to get the language of the Web context which doesn't exist so it fails.

LaurentGoderre:
And the error it gives you is super obscure

wardi:
So, somewhere in the lang() call something is raising a TypeError when we don't have a language for the user? That's obscure, but so is this fix.

Why not pass in `config.get('ckan.locale_default', 'en')` in https://github.com/open-data/ckanext-fluent/blob/master/ckanext/fluent/validators.py#L43 so we're being explicit about wanting to use the default locale, instead of relying on this lang() call failing?

LaurentGoderre:
Ah ok, fair enough

wardi:
I'll fix the lang() failure in this helper as well, let me just restructure the code a bit

Issue #104, opened 2015-12-04
Pagination for organization broken


metaodi:
The pagination for organizations relies on an undocumented method on IGroupForm (see https://github.com/ckan/ckan/issues/2771). Because this method (`group_controller()`) does not exist on `SchemingOrganizationsPlugin`, it falls back to the group controller.

This leads to the generation of links in the pagination like this `/group/org-name?page=2`.

PR follows.

PR #105, opened 2015-12-04
Add group_controller method to make sure correct controller is used

metaodi:
Fix for issue https://github.com/ckan/ckanext-scheming/issues/104


metaodi:
@wardi I don't understand why the build failed, I'm happy to help, but I need a pointer in the right direction. Thanks in advance!

metaodi:
In the meantime https://github.com/ckan/ckan/pull/2779 has been merged, so we should be good to go here.

metaodi:
ping @wardi

PR #113, opened 2016-01-13
fix the order of choices for PR 106

marcfor:
the changes I made in PR 106 didn't render choices in the order
they were listed in the schema when sorted_choices was not set
to true.  this update should fix that


wardi:
fixed by #112

PR #116, opened 2016-01-27
enable tag vocabularies

hvwaldow:
+ provides the property `"vocabulary"`, the value of which is the name of the vocabulary.    
    Example:

 ```javascript
{
   "field_name": "systems",
   "label": "Systems",
   "preset": "multiple_select",
   "vocabulary": "systems",
   "required": true
 },
```

+ provides the property `"free_tags_only"` in the top level of the schema definition. `"free_tags_only": true` will prevent vocabulary tags from being listed in the free tags' display.

+ Currently tested with presets `select` and `multiple_select`.

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -10,6 +10,7 @@ from ckantoolkit import (
     get_converter,
     navl_validate,
     add_template_directory,
+    get_action
 )
 
 from paste.reloader import watch_file
@@ -47,6 +48,9 @@ ignore_missing = get_validator('ignore_missing')
 not_empty = get_validator('not_empty')
 convert_to_extras = get_converter('convert_to_extras')
 convert_from_extras = get_converter('convert_from_extras')
+convert_to_tags = get_converter('convert_to_tags')
+convert_from_tags = get_converter('convert_from_tags')
+free_tags_only = get_converter('free_tags_only')
 
 DEFAULT_PRESETS = 'ckanext.scheming:presets.json'
 
@@ -66,6 +70,7 @@ class _SchemingMixin(object):
     _presets = None
     _template_dir_added = False
     _validators_loaded = False
+    _vocabularies = None
 
     def get_helpers(self):
         if _SchemingMixin._helpers_loaded:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: ckanext/scheming/plugins.py [hunk 1]
```diff
@@ -449,5 +500,6 @@ def _expand_schemas(schemas):
             if fname not in s:
                 continue
             s[fname] = [_expand_preset(f) for f in s[fname]]
+            s[fname] = [_expand_vocabulary(f) for f in s[fname]]
```
wardi: This means vocabulary choices are expanded into the schema once on startup, so the server needs to be restarted if the vocabulary tags are modified?

ckanext/scheming/validation.py
```diff
@@ -27,7 +27,7 @@ def scheming_choices(field, schema):
     """
     Require that one of the field choices values is passed.
     """
-    return OneOf([c['value'] for c in field['choices']])
+    return OneOf([c['value'] for c in field['choices']], testValueList=True)
```
wardi: What does this do? reading the OneOf source it looks like it allows nested lists of values?

ckanext/scheming/plugins.py
```diff
@@ -86,7 +91,7 @@ def get_helpers(self):
             'scheming_field_by_name': helpers.scheming_field_by_name,
             'scheming_get_presets': helpers.scheming_get_presets,
             'scheming_get_preset': helpers.scheming_get_preset,
-            'scheming_get_schema': helpers.scheming_get_schema,
+            'scheming_get_schema': helpers.scheming_get_schema
```
wardi: hey, I like my trailing commas :-)

ckanext/scheming/plugins.py
hvwaldow: That is correct. I figure its better than talking to the database in each request.

ckanext/scheming/validation.py
hvwaldow: Oh, looking again that might be spurious. It was intended to get around the problem that really has to be dealt with here: https://github.com/ckan/ckanext-scheming/pull/116/files#diff-db32db2d8eb1d92a5e5f5ded851dd550R55. But here `c['value']`is always a string it seems.

ckanext/scheming/plugins.py
hvwaldow: Alright, I see its good style and not typo ...

hvwaldow:
This is problematic on a more fundamental level because initialization of a fresh db fails upon trying to `get.vocabulary_list` when there are no vocabularies yet. I am closing to re-think that.

PR #120, opened 2016-02-08
Added a preset for timezone aware datetimes

LaurentGoderre:


ckanext/scheming/helpers.py
```diff
@@ -180,3 +184,93 @@ def scheming_field_by_name(fields, name):
...
+    if timezones == 'all':
+        return to_options(pytz.all_timezones)
+    elif isinstance(timezones, list):
+        return to_options(validate_tz(timezones))
+
+    return to_options(pytz.common_timezones)
+
+# TODO: Move to CKAN core
+def scheming_get_default_timezone():
+    return 'America/Toronto'
```
wardi: wat

ckanext/scheming/tests/test_validation.py
```diff
@@ -235,6 +235,7 @@ def test_datetime_field_rejects_invalid_separate_date(self):
                 type='camel-photos',
                 name='fred_datetime8',
                 a_relevant_datetime_date='31/11/2014',
+                a_relevant_datetime_time=''
```
wardi: that's not what a form sends. why are you changing this?

wardi:
You've added a new validator and form snippet but no new tests. Would you add a field to the camel photos schema that uses this new validator to see that the form snippet works and some tests of the validators to see that they work properly?

ckanext/scheming/helpers.py
LaurentGoderre: Oh damn...I forgot....this is a stub. This should be in CKAN since CKAN has a configuration setting for this but doesn't expose it

ckanext/scheming/tests/test_validation.py
LaurentGoderre: I don't know why but this test started failing without it

LaurentGoderre:
Depends on https://github.com/ckan/ckan/pull/2868

ckanext/scheming/tests/test_validation.py
```diff
@@ -287,10 +282,237 @@ def test_datetime_field_valid_separate_time(self):
...
+                name='fred_datetime_tz1',
+                a_relevant_datetime_tz='31/11/2014',
+            )
+        except ValidationError as e:
+            assert_equals(e.error_dict['a_relevant_datetime_tz'],
+                          ['Date format incorrect'])
+        else:
+            raise AssertionError('ValidationError not raised')
+
+        try:
```
wardi: Can you make these separate tests? Also, if all you're doing is testing the validator, just call the validator in each test, not package_create.

For testing the form see some of the other functional tests.

LaurentGoderre: I am following the example of all the other tests though. All the tests seems to be calling package_create and not the validator directly. Couldn't that be another change?

wardi: The example to follow is TestRequired a little further above. Isn't that more succinct?

wardi: Oh, I'm sorry, your validator depends on having access to the whole data_dict. Carry on (just separate out the separate tests)

LaurentGoderre: Why separate the test if they all are variations of the same test?

wardi: it's a style preference. it's better to have more short test functions than fewer long test functions. With one long function we can't tell what part failed. We would have to go back to the line number in the test file to see what to fix. With separate functions the function that fails tells us what failed just by looking at its name.

LaurentGoderre: That would give long function name to make theme descriptive enough, do you mind?

wardi: I love long test function names!

LaurentGoderre: Perfect then!

ckanext/scheming/tests/test_validation.py
```diff
@@ -287,10 +281,249 @@ def test_datetime_field_valid_separate_time(self):
...
+            raise AssertionError('ValidationError not raised')
+
+    def test_datetime_field_rejects_time_only(self):
+        lc = LocalCKAN()
+        try:
+            lc.action.package_create(
+                type='camel-photos',
+                name='fred_datetime_tz15',
+                a_relevant_datetime_tz_date='',
+                a_relevant_datetime_tz_time='12:35:aa',
```
wardi: this should be a valid time and no date passed at all (no empty string passed), right?

ckanext/scheming/validation.py
```diff
@@ -88,66 +90,118 @@ def validator(key, data, errors, context):
...
+            del extras[inpt]
+
+        if field.get('required'):
+            not_empty(new_key, data, errors, context)
+
+        return (new_key, value)
+
+    date_key, value = get_input('date')
+    value_full = ''
+    if value != '':
```
wardi: why are you comparing to an empty string here?

ckanext/scheming/tests/test_validation.py
LaurentGoderre: Yes

ckanext/scheming/validation.py
wardi: and in three places below?

ckanext/scheming/tests/test_validation.py
LaurentGoderre: Apparently that breaks the test.... :(

wardi: do you mean "reveals a bug"? :-)

ckanext/scheming/validation.py
LaurentGoderre: Yeah, I can probably get rid of that

LaurentGoderre: The problem I have is if I get rid of it, the validator doesn't get called at all it seems

ckanext/scheming/tests/test_validation.py
LaurentGoderre: Not quite since it would have worked, even if the code behind it was suboptimal

LaurentGoderre:
I made the changes to the test and validator

ckanext/scheming/tests/test_validation.py
```diff
@@ -219,15 +225,6 @@ def test_date_field_valid_date_datetime(self):
...
 
-    def test_date_field_valid_date_datetime(self):
-        lc = LocalCKAN()
-        d = lc.action.package_create(
-            type='camel-photos',
-            name='fred_datetime7',
-            a_relevant_datetime=datetime.datetime(2014, 1, 1, 12, 35),
-        )
-        assert_equals(d['a_relevant_datetime'], '2014-01-01T12:35:00')
-
```
wardi: Why have you removed a test for the other datetime field? Why have you changed a number of tests above that test the datetime field?

LaurentGoderre: That was a duplicate test that did the same thing as the previous one (probably a copy/paste problem)

wardi: How about the changes to the other tests above?

PR #121, opened 2016-03-04
Explicity require pytz

LaurentGoderre:
Required for older versions of CKAN that don't include pytz

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 ckanapi
 ckantoolkit>=0.0.2
+pytz
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ setup(
     install_requires=[
         "ckanapi",
         "ckantoolkit>=0.0.2",
+        "pytz",
     ],
     entry_points=\
     """


requirements.txt
```diff
@@ -1,2 +1,3 @@
 ckanapi
 ckantoolkit>=0.0.2
+pytz==2012j
```
wardi: let's make that >= or not specify the version at all

amercader:
+1, just got bitten by this on a production site (my fault for not having ckanext-scheming pinned). Also when are you planning the next release :) ?

amercader:
BTW requirements are also defined on `setup.py`

requirements.txt
LaurentGoderre: I am just following what is in CKAN so that they stay synced

https://github.com/ckan/ckan/blob/master/requirements.txt#L31

LaurentGoderre:
@amercader I added the line to `setup.py` now we just need to agree on the format of that line ;)

amercader:
`pytz>=2012j` seems great to me

LaurentGoderre:
@amercader I talked with @wardi and agreed to probably best to not have specified versions. That way it won't do anything if ckan already specified it

Issue #130, opened 2016-07-21
Add ability to specify a resource to populate domain/range of a metadata field


jqnatividad:
For example, if there is a field called "Dataset Owner", have the ability to point to a resource (e.g. Resource listing all valid city agencies) to validate the field for choices.

By default, if the user does not specify which field columns to use, it will only use the first column.

Otherwise, have the ability to specify which column is the "value" field and which column is the "label" field.

cc @jguo144

wardi:
I'll do this in a couple parts to keep it generic and easy to extend.

First define a scheming field: `"choices_helper": "name_of_helper_function"`. This helper will accept the complete schema and schema field as parameters and be expected to return a choices list in the same format as the scheming field `choices`.

Second add a `scheming_datastore_choices` helper function that will read a datastore resource in the same ckan instance as specified by `"datastore_choices_resource": "my_resource_id_or_alias"` and convert it into a choices list, by default using the first column as the value and second as the label (override with `"datastore_choices_columns": {"value": "my_key_column", "label": "my_label_column"}` ) 

I'm not going to implement any caching initially (because premature optimization and cache invalidation) but that's an easy thing for anyone to implement in their own helper function if necessary.

Choice lists that are too large to be converted to a list of dicts in memory each time they are needed are considered to be out of scope for this change. That sort of thing is possible to deal with but it will need a more complicated interface, and won't be compatible with the existing `select.html`, `multiple_select.html` and `multiple_checkbox.html` form snippets 

Related issues #74, #117
Related earlier work #116

wardi:
done in #132

PR #131, opened 2016-07-22
Choices from a custom helper

wardi:


wardi:
This implements part 1 of #130

PR #132, opened 2016-07-22
h.scheming_datastore_choices

wardi:
Adds the ability to use a datastore table as a choices list to #131


hvwaldow:
Thanks Ian, that is nice!

I wonder whether it is in general a good strategy to use the datastore for CKAN-internal state that should also be applied to similar problems in new extensions.

wardi:
@hvwaldow I think the datastore is great for things like controlled lists. We can have however many columns we want, and it's fast and fairly easy for users to update. Those things aren't true for our tag vocabulary code.

For static lists I've been using external JSON or CSV files and this new interface works with that sort of thing too. I should add helpers for those cases.

hvwaldow:
@wardi OK, thanks. Another question: In case I fear that datastore could still become too slow (e.g. autocomplete for a very large list of terms) and I wanted to cache that in memory, would it be prudent to attach it to app_globals? Or better use Redis?

wardi:
I mentioned in https://github.com/ckan/ckanext-scheming/issues/130#issuecomment-234560768 that this helper won't work for extremely large lists. If your list is too large to simply request the whole list any time you need an item then you'll need a way of asking for completions and translations from individual values->labels. That should work with datastore with the right index but you could do it in redis or by in-memory too.

PR #133, opened 2016-07-27
Update resource_form.html

rbvictor:
When I try to create a dataset with no resources, the error 'You must add at least one data resource' does not appear in the resource_form. When I analyzed lines 656 to 668 in https://github.com/ckan/ckan/blob/master/ckan/controllers/package.py#L656-L668 , I realized that the error dict is empty and only the error_summary is assigned the message. That is why I suggest checking and adding the error summary in this scheming snippet as well. I tried to extend this snippet otherwise, but I didn't succeed. With this patch, I got the error message to reappear on the page.

### DIFF
--- a/ckanext/scheming/templates/scheming/package/snippets/resource_form.html
+++ b/ckanext/scheming/templates/scheming/package/snippets/resource_form.html
@@ -1,4 +1,5 @@
 {% extends 'package/snippets/resource_form.html' %}
+{% import 'macros/form.html' as form %}
 
 {% block errors %}
   {%- if errors -%}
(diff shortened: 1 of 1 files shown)


rbvictor:
I cannot access circleci to check the failed tests, but I only added minor changes.

erlingbo:
The tests fails because of https://github.com/ckan/ckan/pull/3181. This will be fixed when they release 2.5.3

boykoc:
Any chance this can be re-run against a newer clan version? Looks like it was last tried about a year ago against ckan release-v2.3-latest.

PR #140, opened 2016-11-17
Upgrade pytz to match ckan 2.6

Zharktas:
CKAN core upgraded pytz and install will complain about version conflict.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 ckanapi
 ckantoolkit>=0.0.2
-pytz==2012j
+pytz==2016.4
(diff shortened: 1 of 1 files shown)


wardi:
the circleci failures look unrelated. Looks like we're using the wrong version of the schema for testing against older CKANs. I'll take a look at that.

Regarding your change, I think we should make the pytz requirement looser because this extension supports a range of CKAN versions, not just the latest one.

Zharktas:
Apparently since older pytz does not use semantic versioning, it is not possible to specify version ranges. Only option would be specifying just pytz as a requirements

coreyerickson:
I also was encountering this issue deploying scheming with ckan 2.6.0 until removing the pytz version from the requirements file. 

If you installed the scheming requirements to your virtualenv after installing the ckan requirements you will get version conflict.

pytz

wardi:
fixed by #145

PR #143, opened 2017-01-20
add is_required to upload.html

fanjinfei:


fanjinfei:
@wardi So we need to fix form.html in ckan to pass the test here?

wardi:
this just says that the core image_upload macro doesn't have the new parameter you've added. Maybe the default styling of scheming on vanilla ckan makes the URL and image upload sections of the form look like they are required somehow

fanjinfei:
@wardi It seems vanilla ckan need to be fixed. I can not find the styling about 'required'.

wardi:
Ok, I'll close this then

PR #144, opened 2017-01-27
Bump pytz

klikstermkd:
This will make the extension work properly with CKAN 2.6.0 by using the same pytz version with the one installed on CKAN 2.6.0


wardi:
this is a duplicate of #140, I'll take another look at that one's test failures because I prefer to not require a specific pytz (scheming works with many versions of ckan).

PR #146, opened 2017-02-22
Switch to jinja2 macros.

TkTech:
Macros (when used without `with context`) are cached. Includes/snippets/macros[with context]) are not. When displaying many fields on many pages with high read counts this is significant CPU overhead.

This *only* changes *display* templates for minimal user-impact. Any user-defined templates can be updated by wrapping their existing template in:

```python
{%- macro display(h, field, data, errors, entity_type, object_type) -%}
<existing file>
{%- endmacro -%}
```


wardi:
Can't make this change because it breaks absolutely every site that extends the scheming snippets, but I would really like to know what the speedup is my making this change.

PR #148, opened 2017-03-09
fix error text position for multiple_select.html

fanjinfei:


fanjinfei:
ready for review

thriuin:
@wardi Can you review, thanks

wardi:
What's wrong with the error text position? Maybe this should be fixed in the macro instead.

fanjinfei:
The problem is that the error message is put after the help_text (see internal issue 80). The macro input_block has nothing to fix.

wardi:
but if you never pass errors to input_block won't the box be styled incorrectly when there are errors?

fanjinfei:
The part of the multi-select and error inside multi-select.html will end up as {{ caller() }} at https://github.com/open-data/ckanext-wet-boew/blob/master/ckanext/wet_boew/templates/wet_boew/macros/form.html#L286 . It is still inside the input_block.

fanjinfei:
internal issue 80

wardi:
This repo is used outside our site so let's keep the discussion here. Would you look into the test failure and provide a screen shot of the before/after on a bare ckan instance (no wet or custom theme)?

fanjinfei:
The test error is about `curl -fsSL https://raw.githubusercontent.com/michaelcontento/circleci-matrix/master/src/circleci-matrix.sh`. I am not sure what does that mean. For the bare ckan test, what is proper setup so it can use this repo? Any tutor link?

wardi:
@fanjinfei would you merge from master? the tests should be fixed now

fanjinfei:
@wardi still tests failed.

wardi:
you need to fetch and merge from master or rebase again. (your remote is probably be out of date)

PR #161, opened 2017-07-31
Show email address if field.display_email_name_field not defined

mattfullerton:
I'm not sure what field.display_email_name_field is or is or how to use it from a schema, but showing "None" is very confusing, at least for developers who thought they had an email address in there :-)


wardi:
display_email_name_field is just another value you can add to the field in your schema. Should probably be documented, but it's used in the ckan_dataset.json example.

PR #162, opened 2017-09-08
Update form input classes for Bootstrap 3 support

amercader:
This change follows the direct approach of adding the missing classes in all circumstances, both on BS2 and BS3 (ie no selective rendering). It looks mostly identical but it could potentially clash with existing
themes.

For reference here how the form is rendered with the BS3 classes in both versions (the only difference is the select input lengths in BS2 with and without scheming).

WDYT @wardi? Should this be safe enough or is it worth selectively loading depending on the BS version used? (either two full sets of snippets :scream:, or just one set of snippets with a helper and conditionals all over the place? :scream: )

If you are happy with this approach I'll update the resource fields.

### Default theme in Bootstrap 3
localhost-5000-dataset-new 1

### Default theme in Bootstrap 3 + Scheming
localhost-5000-dataset-new

### Default theme in Bootstrap 2
localhost-5000-dataset-new 3

### Default theme in Bootstrap 2 + Scheming
localhost-5000-dataset-new 2

### DIFF
--- a/ckanext/scheming/presets.json
+++ b/ckanext/scheming/presets.json
@@ -9,7 +9,8 @@
         "validators": "if_empty_same_as(name) unicode",
         "form_snippet": "large_text.html",
         "form_attrs": {
-          "data-module": "slug-preview-target"
+          "data-module": "slug-preview-target",
+          "class": "form-control"
         }
       }
     },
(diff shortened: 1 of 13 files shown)


wardi:
Yes, this looks great. If the classes clash with someone's theme they should still be able to override these changes.

amercader:
@wardi Fixed some more inputs and added a snippet for a plain textarea (not markdown)

wardi:
@amercader please ping me when you get a chance to finish this off.

amercader:
@wardi it's ready AFAIC. I'll send the Jason preset separately

PR #163, opened 2017-09-20
Implementation of the json preset

amercader:
This preset allows to input JSON strings, and load these into the dataset object, eg when passing it to a template or via the API.

The JSON fields accepts JSON strings that serialize objects or lists (and nothing else), but also dicts that can be dumped as JSON. eg

```python

call_action('package_create',
            name='test-json',
            'a_json_field'='{"a": 1, "b": 2}')

call_action('package_create',
            name='test-json',
            'a_json_field'={'a': 1, 'b': 2})

```

both create the following dataset:

```json
{
    "name": "test-json",
    "a_json_field": {
        "a": 1,
        "b": 2
    }
}

```

Lists are not allowed directly due to limitations on CKAN validation code.

The JSON fields can also be used in resources.

Tests cover validation, form and display snippets and also some functional tests, both for dataset and resource fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,10 @@ This extension includes the following presets:
   field
 * `"resource_format_autocomplete"` - resource format validation with
   format guessing based on url and autocompleting form field
+* `"json"` - JSON based input. Only JSON objects and arrays are supported.
+  The input JSON will be loaded during output (eg when loading the dataset in
+  a template or via the API).
+
 
 You may add your own presets by adding them to the `scheming.presets`
 configuration setting.
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ This extension includes the following presets:
   field
 * `"resource_format_autocomplete"` - resource format validation with
   format guessing based on url and autocompleting form field
-* `"json"` - JSON based input. Only JSON objects and arrays are supported.
+* `"json_object"` - JSON based input. Only JSON objects are supported.
   The input JSON will be loaded during output (eg when loading the dataset in
   a template or via the API).
 
--- a/ckanext/scheming/helpers.py
+++ b/ckanext/scheming/helpers.py
@@ -339,6 +339,6 @@ def scheming_display_json_value(value, indent=2):
     :rtype: string
     """
     try:
-        return json.dumps(value, indent=indent)
+        return json.dumps(value, indent=indent, sort_keys=True)
     except (TypeError, ValueError):
         return value
--- a/ckanext/scheming/helpers.py
+++ b/ckanext/scheming/helpers.py
@@ -1,17 +1,20 @@
 import re
 import datetime
 import pytz
+import json
 
 from pylons import config
 from pylons.i18n import gettext
 
 from ckanapi import LocalCKAN, NotFound, NotAuthorized
 
+
 def lang():
     # access this function late in case ckan
     # is not set up fully when importing this module
     from ckantoolkit import h
     return h.lang()
+
 
 def scheming_language_text(text, prefer_lang=None):
     """
@@ -324,3 +327,18 @@ def scheming_get_timezones(field):
         return to_options(validate_tz(timezones))
 
     return to_options(pytz.common_timezones)
+
+
+def scheming_display_json_value(value, indent=2):
+    """
+    Returns the object passed serialized as a JSON string.
+
+    :param value: The object to serialize.
+    :returns: The serialized object, or the original value if it could not be
+        serialized.
+    :rtype: string
+    """
+    try:
+        return json.dumps(value, indent=indent)
+    except (TypeError, ValueError):
+        return value
--- a/ckanext/scheming/presets.json
+++ b/ckanext/scheming/presets.json
@@ -113,9 +113,9 @@
       }
     },
     {
-      "preset_name": "json",
+      "preset_name": "json_object",
       "values": {
-        "validators": "ignore_missing scheming_valid_json",
+        "validators": "ignore_missing scheming_valid_json_object",
         "output_validators": "scheming_load_json",
         "form_snippet": "json.html",
         "display_snippet": "json.html"
--- a/ckanext/scheming/presets.json
+++ b/ckanext/scheming/presets.json
@@ -115,7 +115,7 @@
     {
       "preset_name": "json_object",
       "values": {
-        "validators": "ignore_missing scheming_valid_json_object",
+        "validators": "scheming_required scheming_valid_json_object",
         "output_validators": "scheming_load_json",
         "form_snippet": "json.html",
         "display_snippet": "json.html"
--- a/ckanext/scheming/presets.json
+++ b/ckanext/scheming/presets.json
@@ -111,6 +111,16 @@
         "display_snippet": "datetime_tz.html",
         "validators": "scheming_isodatetime_tz convert_to_json_if_datetime"
       }
+    },
+    {
+      "preset_name": "json",
+      "values": {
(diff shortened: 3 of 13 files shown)


amercader:
@wardi this ended up taking a bit more time than expected but is ready for a first review.

README.md
```diff
@@ -224,6 +224,10 @@ This extension includes the following presets:
   field
 * `"resource_format_autocomplete"` - resource format validation with
   format guessing based on url and autocompleting form field
+* `"json"` - JSON based input. Only JSON objects and arrays are supported.
```
wardi: arrays are supported?

ckanext/scheming/helpers.py
```diff
@@ -324,3 +327,18 @@ def validate_tz(list):
...
+    """
+    Returns the object passed serialized as a JSON string.
+
+    :param value: The object to serialize.
+    :returns: The serialized object, or the original value if it could not be
+        serialized.
+    :rtype: string
+    """
+    try:
+        return json.dumps(value, indent=indent)
```
wardi: You'll need sort_keys=True if you want this to be stable.

ckanext/scheming/tests/test_validation.py
```diff
@@ -52,6 +53,8 @@ def test_choice_field_accepts_valid_choice(self):
             )
         assert_equals(d['category'], 'f2hybrid')
 
+        reset_db()
```
wardi: Hmm. Maybe we can move this to to a class based on FunctionalTestBase instead (IIRC reset_db doesn't clean up everything)

ckanext/scheming/tests/test_validation.py
```diff
@@ -534,3 +537,291 @@ def test_invalid_group_type(self):
         p = SchemingGroupsPlugin.instance
         data, errors = p.validate({}, {'type': 'banana'}, {}, 'dataset_show')
         assert_equals(list(errors), ['type'])
+
+
+class TestJSONValidatorsDatasetValid(object):
+
+    def teardown(self):
+        reset_db()
```
wardi: FunctionalTestBase instead, please :-)

README.md
amercader: Arrays are supported, if passed as a JSON string (eg `a_json_field='[{"a", 1}, {"b", 2}]'`), but not if passed directly (eg `a_json_field=[{'a', 1}, {'b', 2}]`) because of how navl handles the validation.

This is controlled at the [validator level](https://github.com/ckan/ckanext-scheming/pull/163/files#diff-f65a08a9d3340b20658060def4af6419R243) though, so it can be easy restricted if you think that might be an issue.

wardi: the asymmetry is a problem: that would mean that you can't call package_patch on a dataset with a list in a json field because package_patch passes the data from package_show (returns a list) to package_update.

1. How about naming this field json_object instead of json and only accept and produce json objects? I can't think of a case where I'd want a metadata field to be either a list or an object.

2. lists can work with simple validators (not navl nested dict ones) I use them with the multiple select fields, ckanext-repeating and ckanext-composite does whole sub-forms with them. Again this might make more sense as a separate json_list preset that only accepts and produces lists. 

If we do support both lists and in one preset I can see someone arguing that we should support the supporting other valid JSON values (`true`, `false`, `null`, floats and strings) but that's ambiguous when also accepting json as a string, so I'd rather not.

amercader: Sounds good, `json_object` it is.

amercader:
@wardi I limited input to JSON objects and renamed to `json_object`. I kept the snippets as `json.html` because they are generic for any JSON, in case more presets are added later.

Let me know if something else is needed.

ckanext/scheming/presets.json
```diff
@@ -113,9 +113,9 @@
       }
     },
     {
-      "preset_name": "json",
+      "preset_name": "json_object",
       "values": {
-        "validators": "ignore_missing scheming_valid_json",
+        "validators": "ignore_missing scheming_valid_json_object",
```
wardi: use scheming_required instead of ignore_missing for the required setting to work on fields using this preset

ckanext/scheming/validation.py
```diff
@@ -240,7 +240,7 @@ def scheming_valid_json(value, context):
         try:
             loaded = json.loads(value)
 
-            if not isinstance(loaded, (dict, list)):
+            if not isinstance(loaded, dict):
                 raise Invalid(
                     _('Unsupported value for JSON field: {}'.format(value))
```
wardi: `.format()` inside `_()`

wardi:
@amercader just those two little fixes I missed last time and this looks good to go!

amercader:
@wardi done!

wardi:
@amercader thank you for this!

sujeetpillai:
Hi,
I'm getting an error when I edit a resource in a dataset that has a json_object field in its schema. Essentially when the 'resource_update' action is fired it also fires a 'package_update' action internally and creates a new revision of the dataset. In that new revision it's trying to push a python dict into the SQLAlchemy update query in the package_extra table rather than converting it to a JSON string. Do you have the same issue?

However there's no issue when I just update the dataset itself. I have a feeling it's related to CKAN not running plugins package_update logic when calling the package_update from within resource_update. 

Thanks,
Sujeet

wardi:
@sujeetpillai I see what happened. We need to convert the value to a JSON string for it to be compatible with package extras. That's missing from the validator so these fields will only work as resources right now.

@amercader do you have time to fix that?

wardi:
This sort of think is required. https://github.com/ckan/ckanext-fluent/blob/master/ckanext/fluent/validators.py#L127-L129 only convert values when no error is present

amercader:
I can't reproduce this manually, and the following test passes for me:

```diff
diff --git a/ckanext/scheming/tests/test_validation.py b/ckanext/scheming/tests/test_validation.py
index c9b61d8..35f9f18 100644
--- a/ckanext/scheming/tests/test_validation.py
+++ b/ckanext/scheming/tests/test_validation.py
@@ -560,6 +560,29 @@ class TestJSONValidatorsDatasetValid(FunctionalTestBase):
 
         assert_equals(dataset['a_json_field'], {'a': 1, 'b': 2})
 
+    def test_valid_json_object_when_editing_resource(self):
+        lc = LocalCKAN()
+        dataset = lc.action.package_create(
+            type='test-schema',
+            name='bob_json_1',
+            a_json_field={'a': 1, 'b': 2},
+            resources=[{
+                'url': 'http://example.com/data.csv',
+                'name': 'a_csv',
+            }],
+        )
+
+        assert_equals(dataset['a_json_field'], {'a': 1, 'b': 2})
+
+        lc.action.resource_update(
+            id=dataset['resources'][0]['id'],
+            name='a_csv_2',
+            url='http://example.com/data.csv',
+        )
+        dataset = lc.action.package_show(id=dataset['id'])
+
+        assert_equals(dataset['a_json_field'], {'a': 1, 'b': 2})
+

```

@sujeetpillai can you share the schema descriptor (json file) that you are using? Or a simplified version that causes the error for you

@wardi I think the validator always returns a JSON string, [here](https://github.com/ckan/ckanext-scheming/blob/d2b7e7b4a7beffe0300a118471477c34c1807671/ckanext/scheming/validation.py#L248) and [here](https://github.com/ckan/ckanext-scheming/blob/d2b7e7b4a7beffe0300a118471477c34c1807671/ckanext/scheming/validation.py#L254). (The last `return value` is never reached and should be removed)

PR #164, opened 2017-09-28
Update resource url validator to match CKAN core

amercader:
Resource URL has been optional for a while

@wardi if someone is using this preset it might change their logic, but at the same time I think it's good to keep presets in sync with core. Perhaps we should release a new scheming version? or is it not worth it?


wardi:
Yes this is overdue.

Sure I'll release a new scheming version when we're sure the json field is working

PR #165, opened 2017-10-10
 Do not dump empty strings in JSON field

amercader:
When passed an empty string (`''`), the helper used to display the fields with JSON objects returned a pair of quotes (`'""'`) because that's what `json.dumps` does. This is what happens when the field is
empty or there is an error in the form and the data is returned.

Added a few tests as well.

### DIFF
--- a/ckanext/scheming/helpers.py
+++ b/ckanext/scheming/helpers.py
@@ -339,8 +339,6 @@ def scheming_display_json_value(value, indent=2):
         serialized.
     :rtype: string
     """
-    if not value:
-        return value
     try:
         return json.dumps(value, indent=indent, sort_keys=True)
     except (TypeError, ValueError):
--- a/ckanext/scheming/helpers.py
+++ b/ckanext/scheming/helpers.py
@@ -338,6 +338,8 @@ def scheming_display_json_value(value, indent=2):
         serialized.
     :rtype: string
     """
+    if not value:
+        return value
     try:
         return json.dumps(value, indent=indent, sort_keys=True)
     except (TypeError, ValueError):
--- a/ckanext/scheming/helpers.py
+++ b/ckanext/scheming/helpers.py
@@ -339,6 +339,8 @@ def scheming_display_json_value(value, indent=2):
         serialized.
     :rtype: string
     """
+    if isinstance(value, basestring):
+        return value
     try:
         return json.dumps(value, indent=indent, sort_keys=True)
     except (TypeError, ValueError):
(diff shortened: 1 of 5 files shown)


amercader:
@wardi WDYT? It's shame I noticed after 1.1.0 was out, but I guess there can always be a 1.1.1 :)

ckanext/scheming/helpers.py
```diff
@@ -339,6 +339,8 @@ def scheming_display_json_value(value, indent=2):
         serialized.
     :rtype: string
     """
+    if not value:
+        return value
```
wardi: I don't think this fits here. the function description would have to become much more complicated. "returns a JSON string, or an empty string, or [] or {} or False or None.." Can we move this check into the snippet instead?

amercader: What about renaming to `scheming_display_json_object` and checking that `isinstance(value, dict)` and if not returning value? This goes on the same direction of having the preset only dealing with objects.

amercader: On second thoughts the snippet is meant to be generic, so I'll do what you suggest and handle it there

amercader:
@wardi ping

ckanext/scheming/helpers.py
```diff
@@ -339,6 +339,8 @@ def scheming_display_json_value(value, indent=2):
         serialized.
     :rtype: string
     """
+    if isinstance(value, basestring):
+        return value
```
wardi: @amercader I still don't understand this special case. When would the helper be passed an already-serialized JSON string, and if we need this special case for some reason it should at least be part of the docstring.

"Returns the object passed serialized as a JSON string with sorted keys and indent given, unless a unicode or bytes value is passed, those are returned unchanged" . This description just leaves me asking "why?? what a strange function!"

ckanext/scheming/helpers.py
```diff
@@ -339,6 +339,8 @@ def scheming_display_json_value(value, indent=2):
...
     try:
         return json.dumps(value, indent=indent, sort_keys=True)
```
wardi: `ensure_ascii=False` is much nicer for text with non-ascii characters

ckanext/scheming/helpers.py
wardi: oh, it's because of form errors, isn't it.. hmm

wardi: If that's the reason you can do `if value is string` in the jinja2 form snippet instead and we don't end up with this odd special case in the helper

PR #166, opened 2017-10-27
Remove check to avoid adding template dir more than once

amercader:
This check prevents the extension from working on CKAN<=2.5. Because of the way plugins/templates are loaded in these versions, the function is called multiple times in different contexts, so after the first pass, the templates folder is not added again eventhough it hasn't been added to the `config` object. See eg:

https://travis-ci.org/frictionlessdata/ckanext-validation/builds/293081805
https://travis-ci.org/frictionlessdata/ckanext-validation/jobs/293081809

We could check the exact key in config with the value to make sure it needs adding but `add_template_folder()` already checks for duplicates so it seems easier to just call it every time.

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -115,9 +115,6 @@ class _SchemingMixin(object):
             }
 
     def _add_template_directory(self, config):
-        if _SchemingMixin._template_dir_added:
-            return
-        _SchemingMixin._template_dir_added = True
         add_template_directory(config, 'templates')
 
     def _load_presets(self, config):


amercader:
@wardi I just realized that CircleCI was also testing 2.5 and 2.4, so I'll investigate why it's failing on my side

amercader:
@wardi I give up. It's a weird combination of CKAN<=2.5 + nose + tests with more than one plugin + running all tests and not just a single file.
I still think it's a harmless change so if you are happy with it it would be great to have it merged. If not I can implement a more thorough check that checks the actual config values.

wardi:
I'm sure this is a bug when modules are added and removed the templates won't get re-added. We should have a better way to do this, but the reason it's here is to prevent adding the template directory twice when two of the plugins are used at the same time.

Maybe just check if it's already present before adding instead?

wardi:
#240 is a better fix

PR #168, opened 2017-11-01
Skip rendering fields in form if form_snippet is false

amercader:
Allow fields to not be displayed in the form if you want to deal with them internally or manually add them to another snippet.

I used `false` instead of `null` to keep the current behaviour of using the text snippet if `form_snippet` was not defined.


amercader:
@wardi let me know if you are happy with this and I'll add docs and tests if so

PR #174, opened 2018-02-27
Don't render delete button on create form

amercader:
Not sure what jinja magic was at play before but this block should only be rendered if `data.id` is present (ie it's an edit). This has surfaced since we starting passing the group_type in the data_dict in ckan/ckan#4032, and is now causing an exception when rendering the new org/group form.


amercader:
@wardi I added tests but they are not run on CircleCI because:

1. Not sure how to run the tests on CircleCI as they are setup now and run two different Solr schemas (pre-2.7 and 2.7).
2. The group_type stuff currently only works against master which is not tested in circle either.

 These tests work locally.

amercader:
@wardi ping

PR #177, opened 2018-04-04
howto for ckanext-scheming

terchris:
Hi
I have a feeling that the ckanext-scheming can help me create additional fields on a organization. But the documentation is just not there. 

So I have decided to write the howto so that others don't need to struggle with this.

### DIFF
--- a/howto
+++ b/howto
@@ -1,0 +1,139 @@
+
+# howto use ckanext-scheming to create custom fields on datasets
+
+It is sometimes useful to add extra fields to organizations, groups and users.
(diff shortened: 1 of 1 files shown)


x-st3v3n:
@terchris Thanks, this looks really useful - it is definitely the clearest and most complete looking documentation for the scheming plugin i've found.  I'm about to install the plugin using CKAN version 2.9.0a.  I sure hope it works.

umarfarooq82:
Thank you.It helped. Kindly share the updated version if you have, with more details

PR #179, opened 2018-04-26
Support custom types in the slug widget

amercader:
Following the same approach as in ckan/ckan#4207 the slug widget takes the dataset/group/org type into account to display the appropriate strings. Fixed some tests.


amercader:
all green now @wardi. c70a44a4151920d58f001738562494635916a624 is needed in master too to get the tests passing in circleci.

PR #182, opened 2018-06-08
Python3 friendly print statements

rossjones:
This tiny PR changes the print statements in command.py to be Py3
friendly


wardi:
fixed by #230

PR #190, opened 2018-08-21
Better check for already loaded helpers

### LINKED ISSUE #183: HelperError: 'scheming_get_dataset_schema' has not been defined
Hi, 

I have been using ckanext-scheming in 2.7.4 CKAN without problems and last week I upgraded to the 2.8 version and I started getting the following error:
The page will load with a "Server Error" and in the logs there is "HelperError: Helper 'scheming_get_dataset_schema' has not been defined."

The error is exactly the same as this one posted by @amercader [here](https://gist.github.com/amercader/42ac78da709cd7a7fe7a7e99a9771e6a)

The behavior is completely random as far as I can percieve. If I press F5 repeatedly there is +- 50% chance of the page loading correctly or the page giving the Error.

I already tried changing the order of the plugins to check if that has any impact and so far nothing is different.

Might this be related with the these issues? 
- [Issue ckan/4310](https://github.com/ckan/ckan/issues/4310) 
- [Issue ckanext-discovery/1](https://github.com/stadt-karlsruhe/ckanext-discovery/issues/1)

Or
Can it have anything to to with [these lines of code](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/plugins.py#L74) that by some reason the helpers get unregistered and they need to be registered again but they do not enter the right if clause?

You guys have any suggestion on what can I address this?

amercader:
Fixes #183 

The root cause of this was ckan/ckan#4414 but it doesn't hurt to have a better check, that will work on CKAN 2.8 for now.

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -7,7 +7,11 @@ import logging
 
 import ckan.plugins as p
 from ckan.common import c
-from ckan.lib.helpers import helper_functions as core_helper_functions
+try:
+    from ckan.lib.helpers import helper_functions as core_helper_functions
+except ImportError:  # CKAN <= 2.5
+    core_helper_functions = None
+
 from ckantoolkit import (
     DefaultDatasetForm,
     DefaultGroupForm,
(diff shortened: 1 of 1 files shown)


amercader:
Mmm, need to make it work across different versions...

Issue #175, opened 2018-03-27
Support For Field Grouping


TkTech:
@wardi thoughts on adding field groups to the core? Here's an example of how they render by default at the moment:

<img width="704" alt="ff" src="https://user-images.githubusercontent.com/72590/37946759-21cedf38-3155-11e8-8977-2f326bacaed4.png">

It makes a few small changes:

- Adds `default_group` as an optional top-level setting on the schema (defaults to "General" if not set)
- Adds `group` as a per-field option, defaulting to the value of `default_group`.
- Adds `group_order` as an optional top-level setting on the schema to specify the display order of groups. `group_order=["General", "Contact", "License"]`

wardi:
sounds awesome! Are you thinking of grouping the edit form the same way?

"group" is pretty overloaded in ckan though, how about "display group" or similar just to reduce possible confusion?

TkTech:
It currently doesn't change the edit form at all, but that's just templating. `display_group` sounds fine to me, I didn't want it to be too long as it's potentially a lot of typing. I'll prepare a PR from our fork.

TkTech:
![image](https://user-images.githubusercontent.com/72590/44009385-82693ea6-9e79-11e8-8033-1e75dd8958c6.png)

- Field groupings now apply to edit pages and resources.
- Groups can have descriptions.
- Styling of the edit page has been redone to be cleaner and use space more effectively.
- Groups can be collapsed by default for large schema pages
- Fields can have default values on presentation, as well as fields with display snippet set to null in which case it's populated after submission but before validation.

TkTech:
This is included in PR #203.

howff:
@TkTech - this looks great! But since #201 seems to have been abandoned how can we get your support for field grouping? Thanks :)

wardi:
@howff until #203 is complete you can always override the `scheming/package/snippets/additional_info.html` template to display fields grouped any way you like in your theme extension.

TkTech:
@howff see the remaining work on #203, which includes these changes (and a lot more) but just needs some help finishing off tests and documentation.

jqnatividad:
@TkTech @wardi just pinging on the status of Field Grouping...

wardi:
I took the required grouping out of #203 to try to keep the changes on that branch smaller. I'd like to see grouping as an optional feature, but it can also be implemented by overriding a couple templates in a theme plugin.

PR #201, opened 2019-01-03
Composite + Repeating As Core Scheming Features

TkTech:
A bit messy but as people were requesting the changes, here they are. Absolutely 100% not-ready-to-merge (no tests!) It's based on PR #176. It tries not to break too many templates which is why there are some real ugly kludges (ex: `scheming_massage_subfield`)

The main changes in this are:

1. Field default values, which can be simple text strings or complex jinja templates. Ex:
```yaml
- field_name: date_published
  label: Date Published
  default: "{{ h.now().strftime('%Y-%m-%d') }}"
  form_snippet: date.html
  validators: require_when_published
```
2. Support for grouping fields, both for display and editing (`display_group: General`) (#175)
3. Resources and packages are rendered identically, simplifying things.
4. The behaviour of ckanext-composite and ckanext-repeating is included in core, but is implemented completely differently. You can have complex, repeating fields and field groups using any snippets, autocompletes, validators, etc... and they'll all "just work". Just provide a list of `subfields` to turn a field into a composite field, and add `repeatable: true` to allow more than one. `form_blanks` can be used to specify how many repeatable fields should be rendered by default. Ex:
```yaml
- field_name: change_history
  label: Change History
  help_text: "A description of changes made to the data since its release. For each change, indicate the date of change and describe what was changed and why. When a dataset is first published, there is one change history entry for the initial creation of the dataset.  Add additional change history entries as needed."
  display_group: Science
  repeatable: true
  form_blanks: 1
  subfields:
  - field_name: change_description
    label: Description of Change
    default: "Initial creation of dataset record"
    form_snippet: textarea.html
    display_snippet: text.html
    validators: require_when_published

  - field_name: change_date
    label: Date of Change
    form_snippet: date.html
    display_snippet: date.html
    default: "{{ h.now().strftime('%Y-%m-%d') }}"
```
5. Missing call to the help text snippet is added to slug.html

@wardi requested you as a reviewer just to get your thoughts if you have time. I will likely squash this and redo it before anything gets merged so feedback is appreciated.


circle.yml
```diff
@@ -20,7 +20,6 @@ dependencies:
   override:
   - git clone https://github.com/ckan/ckan
   - python setup.py develop
-  - pip install -r requirements.txt
```
wardi: why was this removed?

ckanext/scheming/commands.py
```diff
@@ -1,65 +0,0 @@
...
-                field_names = ('dataset_fields', 'fields', 'resource_fields')
-
-                for field_name in field_names:
-                    if s[typ].get(field_name):
-                        if field_name == 'resource_fields':
-                            print " * " + json.dumps("resource")
-                        for field in s[typ][field_name]:
-                            print "   - " + json.dumps(field['field_name']),
-                            print scheming_language_text(field.get('label'))
-            print
```
wardi: removing to reduce paste dependencies?

circle.yml
wardi: ah, moved to install_requires

wardi:
moving the examples into their own directory is good, but our readme says to use  `scheming.dataset_schemas = ckanext.scheming:camel_photos.json` which will no longer work. Could move the examples dir to ckanext/scheming/examples instead so that this type of import will still work

ckanext/scheming/plugins.py
```diff
@@ -69,65 +51,68 @@ class _SchemingMixin(object):
...
+                'scheming.presets',
+                DEFAULT_PRESETS
+            ).split()
+        )
+
+        _SchemingMixin._presets = {
+            field['preset_name']: field['values']
+            for preset_path in presets
+            for field in _load_schema(preset_path)['presets']
+        }
```
wardi: I like renaming `f` to `preset_path` but otherwise I find the original code easier to understand

ckanext/scheming/plugins.py
```diff
@@ -234,28 +220,44 @@ def validate(self, context, data_dict, schema, action):
...
-        if action_type == 'show':
-            get_validators = _field_output_validators
-        elif action_type == 'create':
-            get_validators = _field_create_validators
-        else:
-            get_validators = _field_validators
+        get_validators = {
+            'show': _field_output_validators,
+            'create': _field_create_validators,
+            'upate': _field_create_validators
```
wardi: `upate`?  Also update doesn't use the `_field_create_validators`. I prefer the original code here even if it's longer.

ckanext/scheming/plugins.py
```diff
@@ -234,28 +220,44 @@ def validate(self, context, data_dict, schema, action):
...
+                )
+
+                # Apply default field values before going through validation. This
+                # deals with fields that have form_snippet set to null, and fields
+                # that have defaults added after initial creation.
+                if data_dict.get(f['field_name']) is None:
+                    default = f.get('default')
+                    if default:
+                        data_dict[f['field_name']] = (
+                            helpers.scheming_render_from_string(
```
wardi: I love jinja template defaults, but let's name them "default_template" instead of "default" so we still have the fast path for common non-jinja defaults.

ckanext/scheming/plugins.py
```diff
@@ -362,10 +364,12 @@ def _load_schema_module_path(url):
...
-    p = os.path.join(os.path.dirname(inspect.getfile(m)), file_name)
-    if os.path.exists(p):
-        watch_file(p)
-        return loader.load(open(p))
+
+    path = os.path.join(os.path.dirname(inspect.getfile(m)), file_name)
+    if os.path.exists(path):
+        watch_file(path)
+        with open(path, 'rb') as schema_file:
+            return yaml.load(schema_file)
```
wardi: I don't love parsing json with yaml because then users are going to sneak yamlisms into their schemas without renaming them. Also since pyyaml has code execution vulnerabilities this opens up sites with live schema editors to potential problems

ckanext/scheming/plugins.py
```diff
@@ -412,16 +415,25 @@ def _field_validators(f, schema, convert_extras):
...
+            f['validators'],
+            f,
+            schema
+        )
     elif helpers.scheming_field_required(f):
-        validators = [not_empty, unicode]
+        validators = [not_empty]
     else:
-        validators = [ignore_missing, unicode]
+        validators = [ignore_missing]
```
wardi: huh. I wonder why I had unicode there. Good change.

ckanext/scheming/helpers.py
```diff
@@ -343,3 +347,46 @@ def scheming_display_json_value(value, indent=2):
...
+
+    return template.render(**kwargs)
+
+
+def scheming_massage_subfield(field, subfield, index, data):
+    # Subfield & data are re-used, we must not modify the original!
+    sf = subfield.copy()
+    data = data.copy()
+
+    new_field_name = '{field_name}-{index}-{subfield_name}'.format(
```
wardi: I learned that our controllers already have a convention for nested parameters: 
https://github.com/ckan/ckan/blob/2.7/ckan/controllers/package.py#L564-L565

This converts parameters in the form `field_name__index__subfield_name` into lists of nested dicts automatically in the controller. We should use this instead of inventing another nested list-of-dicts implementation for scheming.

ckanext/scheming/templates/scheming/form_snippets/_organization_select.html
```diff
@@ -10,7 +10,7 @@
     error=errors[field.field_name],
     is_required=org_required,
     classes=['form-group', 'control-medium'],
-    extra_html=caller() if caller,
```
wardi: trailing commas are my right as a python programmer gosh darn it! :-)

ckanext/scheming/templates/scheming/form_snippets/slug.html
```diff
@@ -32,4 +32,4 @@
     is_required=h.scheming_field_required(field)
     ) }}
 
-
+{%- snippet 'scheming/form_snippets/help_text.html', field=field -%}
```
wardi: was this getting repeated?

ckanext/scheming/templates/scheming/package/resource_read.html
```diff
@@ -9,58 +9,61 @@
...
+                            object_type=dataset_type -%}
+                                </dd>
+                            {% endfor %}
+                        </dl>
+                    </div>
+                {% endif %}
+            {% endfor %}
+        {% endblock %}
+
+    </div>
```
wardi: ckan's templates are indented 2 spaces, changing all the indentation to 4 is making it harder to see the actual change here

setup.py
```diff
@@ -19,17 +15,17 @@
...
+        'pytz'
+    ],
     entry_points=\
     """
     [ckan.plugins]
     scheming_datasets=ckanext.scheming.plugins:SchemingDatasetsPlugin
     scheming_groups=ckanext.scheming.plugins:SchemingGroupsPlugin
     scheming_organizations=ckanext.scheming.plugins:SchemingOrganizationsPlugin
-    scheming_test_subclass=ckanext.scheming.tests.plugins:SchemingTestSubclass
-    scheming_test_plugin=ckanext.scheming.tests.plugins:SchemingTestSchemaPlugin
```
wardi: removing test plugins?

ckanext/scheming/helpers.py
wardi: This could be too large a change to ask for as part of getting this PR merged. Let's just revisit later.

wardi:
This PR just needs conflicts resolved and some minor fixes mentioned above.

ckanext/scheming/templates/scheming/form_snippets/slug.html
TkTech: It was added not removed, no help text was being displayed for slugs.

ckanext/scheming/helpers.py
TkTech: It's commented elsewhere, but it's done this way because that's already how composite and repeating do it. Without this, plugins that were built on top of them cannot find the fields.

ckanext/scheming/commands.py
TkTech: Needs paste and an update for py3. Didn't seem terribly useful so with limited time I just removed it.

ckanext/scheming/plugins.py
TkTech: Which takes precedence when overloading from presets?

ckanext/scheming/plugins.py
TkTech: There was a reason for this, to make the assignment a single atomic operation...can't remember why tho, I think something I was experimenting with in load_schema would call something else that would in turn look at the presets again...

ckanext/scheming/plugins.py
wardi: let's say default_template since it's cooler. If there's a default_template in a preset but a schema wants a static value they can either use the same static value as the new default_template or set `default_template: null`

ckanext/scheming/plugins.py
wardi: ok, so we could build this as a new dict then overwrite the shared one to make it atomic.

The way these values are cached in class variables probably isn't safe even with the atomic update for example with tests reloading configurations. Need a better way to store and update these.

ckanext/scheming/commands.py
wardi: I'm fine with this

wardi:
closing in favour of #203

ckanext/scheming/templates/scheming/package/resource_read.html
```diff
@@ -9,58 +9,61 @@
...
+                    <th scope="row">{{ _('License') }}</th>
+                    <td>{% snippet "snippets/license.html", pkg_dict=pkg, text_only=True %}</td>
+                </tr>
+            {%- endblock -%}
+            </tbody>
+        </table>
+
+        {% block resource_fields %}
+            {%- for group in schema.resource_fields|groupby('display_group') -%}
+                {%- set group_fields = h.non_empty_fields(group.list|list, res, exclude_fields) -%}
```
wardi: Where's this defined?

TkTech:
```
def non_empty_fields(field_list, pkg_dict, exclude):
    r = []
    for field in field_list:
        if field['field_name'] in exclude:
            continue

        if field.get('display_snippet', False) is None:
            continue

        if pkg_dict.get(field['field_name']):
            r.append(field)
return r
```

Was missed when pulling chunks out for the PR.

PR #203, opened 2019-02-15
Composite + Repeating integration [WIP]

wardi:
applied changes suggested in my review of #201, new branch on this repo so we can collaborate


LukeTully:
@wardi I've been working on a hacked together integration for these features and might be able to contribute. Is there a roadmap for this or do you have a sense of what is remaining?

wardi:
Just needs some fixes and some new tests and docs for the changes, are you looking to help with this PR?

LukeTully:
Ya, I came across this as I was working on a hacked together solution for a new complex schema, and would rather devote that energy to this. Just made the comment in passing though, haven’t looked through how this implementation differs from a combo of the repeating and composite plugins.

PR #210, opened 2019-06-19
Preliminary support for validators in dict WIP

### LINKED ISSUE #75: Support list of dicts
I want to accept arbitrary dicts for a field that i defined in a scheme. It seems that CKAN tries to apply some kind of sub-schema if it encounters a list of dicts. Is it possible to define such sub-schemas with this extension?
And if the fields are not known, all the data is [put into the key "__junk"](https://github.com/ckan/ckan/blob/0f145a54629287a6f7764b8fd73963bda62260ef/ckan/lib/navl/dictization_functions.py#L147-L149) as a flattened dict.

What I ended up doing is writing my own validator that gets the values out of the __junk, but I'm pretty sure this is a bad solution (see here for details: https://github.com/ogdch/ckanext-switzerland/blob/421de6c546462f0ac5764f93064d68d285addda2/ckanext/switzerland/validators.py#L39-L62).

Any ideas how to handle this?

Zharktas:
Fixes #75 

This PR enables to have dict of validators instead of just list of string.

For example

```
    {
      "field_name": "groups",
      "label": "Categories",
      "preset": "multiple_checkbox",
      "validators": {
        "id": "ignore_missing unicode_safe",
        "name": "ignore_missing unicode_safe",
        "title": "ignore_missing unicode_safe",
        "__extras": "ignore"
      },
      "output_validators": {
        "id": "ignore_missing",
        "name": "ignore_missing",
        "title": "ignore_missing"
      }
    }
```

If the schema defines groups (which is a core field) in any kind of definition, posting groups with the dataset eg. using package_create in the api will fail with validation error of having things in __junk.

This is work in progress as it probably has some things not yet considered and missing tests, but i wanted to put it out for discussion anyway.

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -30,6 +30,7 @@ from ckanext.scheming import loader
 from ckanext.scheming.errors import SchemingException
 from ckanext.scheming.validation import (
     validators_from_string,
+    validators_from_dict,
     scheming_choices,
     scheming_required,
     scheming_multiple_choice,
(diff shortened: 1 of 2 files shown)


Zharktas:
Just realized that it is not possible in ckan itself that some validators are applied to field and then subschema is used to that same field. e.g. add groups to dataset form. So I'm closing this as groups need some work around anyway.

Issue #219, opened 2019-11-06
form-control class creates two inputs fields, one above the other


frafra:
Removing `form-control` fixes the issue:
https://github.com/ckan/ckanext-scheming/blob/1a096dd963424884bbaf70fa28da5c1de058d27d/ckanext/scheming/templates/scheming/form_snippets/multiple_select.html#L29

![Screenshot_2019-11-06 Datasets - COAT Data Portal](https://user-images.githubusercontent.com/4068/68318174-a3ca0980-00bc-11ea-879e-ad7c823c842e.png)

frafra:
Same glitch happens for file format.

howff:
Are you sure it's multiple_select.html which renders the tags(keywords) in your screenshot?
I see the same problem when using tags but multiple_select.html is not used, text.html is used.
However editing text.html to remove form-control affects all normal text entry fields.
The solution seems to be adding a new form_snippet, say tags.html, which is a copy of text.html but without the form-control class, and referencing this form_snippet in the preset.

(No comment on the file format issue, I've not tried it yet)

frafra:
I found this problem in tags first, but I had to fix it on `text.html` too.

bstutsky:
This also seems to be an issue for `resource_format_autocomplete`

LukeTully:
@frafra What did your schema look like for this? Did you resolve the issue?

frafra:
@LukeTully I just removed the class from the template manually as workaround.

Patrick1Rhode:
Note guys, we also have experienced this issue.

![image_2021_04_16T16_08_44_961Z](https://user-images.githubusercontent.com/8822597/115056781-87eaaa00-9ee3-11eb-871a-f8ac56f9fa47.png)


We have solved this and other problems by adding a custom CSS.

.form-control {
    padding: unset;
}

.control-full input, .control-full select, .control-full textarea {
    height: 34px;
}

.select2-container{
    height:auto;
}

Can we evaluate this CSS in the core?

frafra:
This bug has been fixed.

PR #225, opened 2019-11-20
tags needs its own form_snippet to fix css class

howff:
CSS includes form-control which causes incorrect layout.
In order to remove form-control we need a new form_snippet for tags; same as for text but without form-control.
Fixes https://github.com/ckan/ckanext-scheming/issues/219
except only for tags not file format or other glitches.

### DIFF
--- a/ckanext/scheming/presets.json
+++ b/ckanext/scheming/presets.json
@@ -24,6 +24,7 @@
       "preset_name": "tag_string_autocomplete",
       "values": {
         "validators": "ignore_missing tag_string_convert",
+        "form_snippet": "tags.html",
         "classes": ["control-full"],
         "form_attrs": {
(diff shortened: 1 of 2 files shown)


LukeTully:
@howff Did you find why this was? Is form-control used as the selector for the tag_string_autocomplete preset javascript? Seems like adding a class called "NOT-form-control" could be approached differently if we know what specifically is causing the issue.

howff:
I didn't try too hard to find the exact cause, sorry it was a quick hack to get it working. I left the CSS in there and added the NOT prefix just to indicate where the fault lies, not because I want a class with that  name. The effect seemed to be caused by the class appearing twice, once on an element which is underneath the element we want, and having the class applied to the top one caused it to be offset from the one underneath instead of hiding it.

Issue #14, opened 2014-12-16
Feature request: retain extra fields


florianm:
Not sure whether this is a bug or a feature, but it seems that enabling ckanext-scheming with a custom schema will, on first saving a dataset with the new schema, delete any existing "extra" fields. Even to only retain the existing extra fields to safely migrate data would be helpful - this could be achieved if config settings would toggle the display of existing extra fields and/or form fields.

florianm:
@wardi as scheming disables free-form extra fields, and harvesting requires them, what's the path of least resistance to ckan_harvest between two ckan instances with (easy: the same, harder: different) custom dataset schemas?

wardi:
@florianm free-form extras should work if we just add the edit fields back to the template. Care to give that a try?

The only problem is: how do we handle free-form extras with names that match the ones in the schema?

CarlQLange:
Just in case someone wants this (and I think lots do), here's a copy-pastable solution:

In `ckanext-scheming/ckanext/scheming/templates/scheming/package/snippets/package_form.html`

from

```html
{% block metadata_fields %}
{% endblock %}
```

to

```html
{% block metadata_fields %}
  {% block package_metadata_fields_custom %}
    {% block custom_fields %}
      {% snippet 'snippets/custom_form_fields.html', extras=data.extras, errors=errors, limit=3 %}
    {% endblock %}
  {% endblock %}
{% endblock %}
```

NicolaiMogensen:
@CarlQLange Thanks a lot, this works just fine (y)

BonaBeavis:
@wardi There is already logic which checks if a key is defined in the root or under extras.

PR #229, opened 2019-12-10
Retain extra metadata fields

howff:
Allow users creating datasets to enter additional metadata, as per the default when scheming is not used. 

Fixes

https://github.com/ckan/ckanext-scheming/issues/14


wardi:
This would have to be configurable in the schema. Existing users of scheming won't want these fields to reappear suddenly (and users have to be careful about not using conflicting field names)

PR #230, opened 2019-12-12
[WIP]  Py3

smotornyuk:
Just changes mentioned in [Migration guide](https://github.com/ckan/ckan/wiki/Migration-guide-for-extensions#step-by-step-workflow). For now, the only CLI command exposed as separate entry-point - `ckanext-scheming show`. But after https://github.com/ckan/ckan/pull/5112 i'm going to use `IClick` instead. In addition, there will be some changes to tests, after pytest adoption.


bin/travis-build.bash
```diff
@@ -21,7 +21,7 @@ fi
 
 python setup.py develop
 
-if [ -f requirements-py2.txt ]
+if [ -f requirements-py2.txt ] && [ $CKANVERSION != 'master' ]
```
wardi: this should be testing the python version not the ckan version, right?

ckanext/scheming/commands.py
```diff
@@ -25,41 +23,17 @@ class SchemingCommand(CkanCommand):
...
-
-                for field_name in field_names:
-                    if s[typ].get(field_name):
-                        if field_name == 'resource_fields':
-                            print " * " + json.dumps("resource")
-                        for field in s[typ][field_name]:
-                            print "   - " + json.dumps(field['field_name']),
-                            print scheming_language_text(field.get('label'))
-            print
+            print(self.__doc__)
```
wardi: I'm fine with removing the scheming cli instead; it does nothing that you can't already do with the API now.

ckanext/scheming/helpers.py
```diff
@@ -275,7 +280,7 @@ def date_tz_str_to_datetime(date_str):
         microseconds = int(m.groupdict(0).get('microseconds'))
         time_tuple = time_tuple[:5] + [seconds, microseconds]
 
-    final_date = datetime.datetime(*map(int, time_tuple))
+    final_date = datetime.datetime(*list(map(int, time_tuple)))
```
wardi: this `list()` is unnecessary. Even better would be
```python
final_date = datetime.datetime(*(int(x) for x in time_tuple))
```
then we could remove the `import map` above

ckanext/scheming/plugins.py
```diff
@@ -22,8 +25,6 @@
     add_template_directory,
 )
 
-from paste.reloader import watch_file
```
wardi: @smotornyuk this was the hook I was talking about on Tuesday. We need one of these on flask. Would be a good candidate for addition to ckantoolkit

ckanext/scheming/plugins.py
```diff
@@ -426,9 +439,9 @@ def _field_validators(f, schema, convert_extras):
     if 'validators' in f:
         validators = validators_from_string(f['validators'], f, schema)
     elif helpers.scheming_field_required(f):
-        validators = [not_empty, unicode]
+        validators = [not_empty, str]
```
wardi: `unicode_safe` would be a better choice than `str`.

ckanext/scheming/templates/scheming/form_snippets/license.html
```diff
@@ -9,7 +9,7 @@
 {%- if field.get('sorted_choices', true) -%}
     {%- set licenses = licenses|sort -%}
 {%- endif -%}
-{%- for license_desc, license_id in licenses -%}
+{%- for license_id, license_desc in licenses -%}
```
wardi: is this just a separate bug fix?

ckanext/scheming/tests/mock_pylons_request.py
```diff
@@ -1,3 +1,4 @@
+from builtins import object
```
wardi: what's this for?

ckanext/scheming/validation.py
```diff
@@ -310,7 +311,7 @@ def get_validator_or_converter(name):
     Get a validator or converter by name
     """
     if name == 'unicode':
-        return unicode
+        return str
```
wardi: `six.text_type`?

wardi:
@smotornyuk looks great so far!

ckanext/scheming/templates/scheming/form_snippets/license.html
smotornyuk: Yes,  we are no longer passing licenses as `c.attribute` with `dataset` Blueprint - but this change shouldn't be here, because it's not compatible with older CKAN versions. I'll remove it

ckanext/scheming/tests/mock_pylons_request.py
smotornyuk: It's auto-fix provided by the `future` library - it provides exactly the same `object` for python2 as the one from python3. I don't think that we are relying on some low-level behavioral patterns of `object` that may be changed in python3, so it's likely to be unnecessary.

bin/travis-build.bash
smotornyuk: Completely agree

ckanext/scheming/commands.py
smotornyuk: The less code, the better

ckanext/scheming/plugins.py
smotornyuk: I haven't yet a chance to check werkzeug's observers, but it seems that they are not exposing any handles to the outside world.

While `paste` has this `watch_file`, that internally call function, that reloads server, `Flask` requires the complete list of watched files that are registered during server startup. If it's true than having something like `watch_file` inside `ckantoolkit` can cause a lot of problems - it must be called only on the initialization stage, not in runtime, but people definitely will call it everywhere. In that case, it would be safer to use only paths\globs specified statically in the config file and don't introduce any programmatical way to control server reloading.

Theoretically, we can patch `config` object and touch any source file\config file - thus Flask's observer will trigger reload, thinking that config\code was modified, and will fetch all the changes, but it's too tricky solution for keeping it in code.

smotornyuk: PS I hadn't completely removed this watcher - there is an attempt to use it in on L380. So, existing CKAN versions still support reloading and we still have time to include solution into the upcoming version

bin/travis-build.bash
```diff
@@ -41,7 +48,12 @@ sudo -u postgres psql -c 'CREATE DATABASE ckan_test WITH OWNER ckan_default;'
 
 echo "Initialising the database..."
 cd ckan
-paster db init -c test-core.ini
+if [ $CKANVERSION == 'master' ]
+then
+    ckan -c test-core.ini db init
```
wardi: should be checking > '2.8' instead

ckanext/scheming/plugins.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 # encoding: utf-8
 
+from future import standard_library
+standard_library.install_aliases()
```
wardi: is this necessary? `install_aliases` does a ton of monkey-patching and module re-loading, would be nice to avoid if we can.

smotornyuk: it enables `urllib` alias for python2 - https://github.com/ckan/ckanext-scheming/pull/230/files/696c4295343925925bb18bfe580c04714a0c8fd4#diff-83fc01bdddb5a3e63073f84bc5f67fa1R388

I'll change it to `six` version  - looking on it from the perspective of all migrated extension, now i see, that it probably would be better not to use `future` library as almost everything available in six or via some polyfill.

wardi: Thanks! When I was testing install_aliases in another package it seemed like it caused more problems than it solved.

bin/travis-build.bash
smotornyuk: Good point. I'll use `2.9` instead, as in 2.8 we were mainly using paster. For the consistency, I'll leave single bracket there, but it will result in a bit ugly condition block:
`if [ $CKANVERSION \< '2.9' ]...`
In this case, `<` must be escaped, otherwise, it will work as input redirect. double-brackets would solve the problem, but we are using single-brackets in all the extensions and I'm not so confident in my sh knowledge to guarantee, that switching into a bit different command won't create unexpected side-effects.

ckanext/scheming/plugins.py
smotornyuk: Unfortunately, i've just got another proof of your words. [This issue in ckanext-dcat](https://github.com/ckan/ckanext-dcat/issues/179) was caused by using `future.install_aliases` :sob: 
Dev from our team was going to create PR with fix into ckanext-dcat, and I have to update all my pending PRs into extensions, removing `future`.

smotornyuk:
Yay, green tests for public versions. tomorrow I'm going to start working on adapting nosetests to pytest format

smotornyuk:
Tests on the master are passing with python 2. @wardi , Should I configure Travis for running tests on python 3, or you prefer waiting till CKAN 2.9 public announce?

smotornyuk:
@wardi , ready for review

amercader:
Woo, great work @smotornyuk

test-requirements.txt
```diff
@@ -1,2 +1,3 @@
 factory-boy==2.1.1
 mock==2.0.0
+bs4
```
wardi: looks like the correct name for this package is beautifulsoup4

https://pypi.org/project/bs4/

PR #233, opened 2019-12-23
Improvement: Add support for organization image file uploads.

boykoc:
This PR creates a new form template and preset to handle Organization logo file uploads. 

After reviewing core CKAN templates I noticed some differences in some variables and calling of the `form.image_upload` macro between resources and organization uploads.

**Current Behaviour:**
Using the existing `upload.html` template and related preset does not properly upload organization logos. The file name is not put through the validators, and the frontend is not able to locate the file on the server ( e.g. `adsffffdsfddsf`). However, there does not seem to be any errors thrown. The form submits successfully and the image display name is saved as the file name (before validation).

**Expected Behaviour:**
User should be able to upload an image to an organization, have it's validators run, and be visible on the frontend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,8 @@ This extension includes the following presets:
 * `"dataset_organization"` - organization validation and form select box
 * `"resource_url_upload"` - resource url validaton and link/upload form
   field
+  * `"organization_url_upload"` - organization url validaton and link/upload form
+  field
 * `"resource_format_autocomplete"` - resource format validation with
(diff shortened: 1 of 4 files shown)


wardi:
@boykoc thank you!

Would you be able resolve the conflict and update the corresponding test under tests/nose (we now have two copies of all our tests for ckan master/python3 compatibility)

boykoc:
@wardi sure thing

PR #241, opened 2020-04-14
Restore licenses in v2.9

smotornyuk:
In CKAN 2.9 license options for the dataset form are generated [in template](https://github.com/ckan/ckan/blob/master/ckan/templates/package/snippets/package_basic_fields.html#L36) and there is no `c.licenses`. Helper function in v2.9 generate slightly different result from what we have in previous versions, so it would be safer to reproduce old behavior and add `c.licenses` via `setup_template_variables`. Personally, I'd like to use logic from the master, but it will break existing overrides to `license.html` . So, it can't be helped and we must add `license` into `c` object.

### DIFF
--- a/ckanext/scheming/tests/test_form.py
+++ b/ckanext/scheming/tests/test_form.py
@@ -86,6 +86,18 @@ class TestDatasetFormNew(object):
         form = BeautifulSoup(response.body).select_one("#resource-edit")
         assert form.select("input[name=camels_in_photo]")
 
+    @pytest.mark.ckan_config("scheming.dataset_schemas", "ckanext.scheming:ckan_dataset.json")
+    def test_dataset_form_includes_licenses(self, app, sysadmin_env):
+        """Starting from CKAN v2.9, licenses are not available as template
+        variable and we are extendisn
+        `DefaultDatasetForm::setup_template_variables` in order to change
+        it.
+        """
+        response = app.get(url="/dataset/new", extra_environ=sysadmin_env)
+        page = BeautifulSoup(response.body)
+        licenses = page.select('#field-license_id option')
+        assert licenses
+
 
 @pytest.mark.usefixtures("clean_db")
 class TestOrganizationFormNew(object):
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -9,6 +9,7 @@ from functools import wraps
 import six
 
 import ckan.plugins as p
+import ckan.model as model
 from ckan.common import c
 try:
     from ckan.lib.helpers import helper_functions as core_helper_functions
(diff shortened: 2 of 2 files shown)


ckanext/scheming/tests/test_form.py
```diff
@@ -86,6 +86,18 @@ def test_resource_form_includes_custom_fields(self, app, sysadmin_env):
         form = BeautifulSoup(response.body).select_one("#resource-edit")
         assert form.select("input[name=camels_in_photo]")
 
+    @pytest.mark.ckan_config("scheming.dataset_schemas", "ckanext.scheming:ckan_dataset.json")
```
wardi, line 89: @smotornyuk this line has a side effect that seems to remove this settings from the ini for later tests https://travis-ci.org/github/ckan/ckanext-scheming/jobs/677860586 . Seems like a bug..

smotornyuk: Not exactly. Tests must be as independent as possible. But there are a few places in CKAN, which make "pure" testing a bit difficult. Most noticeable are singleton( model.Session, g, etc.) and memoization(`logic.get_action` memoizes all the actions, similar things done for auth functions and licenses). 

In this particular case, you are using `LocalCKAN` in `TestDatasetSchemaLists::test_dataset_schema_list`, which just calls `logic.get_action` internally. `get_action` reuses existing actions cache, left by the previous test, with only `ckan_dataset` schema enabled. As all the actions are cached and you have to explicitly rebuild this cache in every separate test(or use `environment.update_config`). Otherwise, it won't be a "pure" test. That's why I'm usually using the `app` fixture - it automatically recreating the test application for each test. 

And if you change the following:
```python
    def test_dataset_schema_list(self):
```
to
```python
    def test_dataset_schema_list(self, app):
```
the test will pass because the `app` fixture recreates the test app and build a completely new action cache, correct for the current list of enabled schemas.

smotornyuk: PS there are two common strategies for making tests "pure" - **setup** before the actual test, or **teardown** after the test. Yes, you can do both, but it's usually required only if you definitely need to do some clean-up(remove temporary files). 
Personally I prefer to rely on **setup** - you just preparing everything before the test is executed, so it will give you consistent results, no matter what was done before the test. And you can check artifacts(DB records, files) when the test is passed/failed if there is no clean-up afterward.
On the other hand, if you use **teardown**, you are removing garbage after the test, but in the beginning, there is no clean-up and, in such a way, you are expecting that previous test was successfully cleaned everything - this makes tests coupled and less "pure".

As I said, **setup** and **teardown** may be used together. But for making tests independent, only first is required, and usually its more than enough. For example, there is not much sense in rebuilding the database BEFORE and once again AFTER tests. If you running tests inside the container, it will be dropped anyway. If you running tests locally, DB will be rebuilt the next time, as a part of **setup** in consequent test runs.

wardi: thanks @smotornyuk that explains it

PR #242, opened 2020-04-18
pytest everything

wardi:
use pytest for all supported versions of ckan


shevron:
Hi, since this is extremely useful to me I created a stand-alone pytest plugin that wraps your code and can be used on any CKAN extension: https://github.com/datopian/pytest-ckan . 

Would love to hear any comnents or feedback.

Issue #66, opened 2015-07-30
Best way to hide a field in the Additional table


amercader:
I know this is not supported, but what would be the best way to define on your schema that you don't want a field to be displayed on the Additional Info table?

[This check](https://github.com/open-data/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/package/snippets/additional_info.html#L16) led me to try something like:

```json
    {     
      "field_name": "private",    
      "label": "Visibility",    
      "required": true,    
      "validators": "boolean_validator datasets_with_no_organization_cannot_be_private opendatani_private_datasets",
      "form_snippet": "visibility_form_field.html",
      "display_snippet": null
    },
```

But it had no effect.

Arguably the `private` dataset should be one of the ones [excluded](https://github.com/open-data/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/package/snippets/additional_info.html#L3), but I'm thinking more about a generic option.

What do you think about `"display_snippet": null` or a new explicit `"display": false` key?

wardi:
If `"display_snippet": null` isn't working then that's a bug. `"form_snippet": null` is how we hide form snippets and I'd like display_snippet to behave the same way.

If private is one that should be excluded by default let's fix that too. I'm not very happy with how the excluded list is defined in the template right now. I think it would be better to have some default "display_snippet": none values for specific fields in the schema instead (so it's easy to override or add more excluded items)

Zharktas:
@wardi apparently resource fields do not have hiding with display_snippet: null

https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/package/snippets/additional_info.html#L16

https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/package/resource_read.html#L48

Although it would be easy to fix, I also think that this feature should be documented.

ryangermann-gov-on-ca:
What might be generally useful would be a parameter that lets Scheming fields be set to "private" so that they only render under the same conditions as Form Profiles will be discoverable: if a user is a mamber of an Organization, they can see Private Dataset pages, but also for Public datasets otherwise discoverable by non-authenticated users, they would see "private" Additional fields.

PR #243, opened 2020-05-07
wip: Add validator for null form_snippets to prevent removing values.

boykoc:
When using ckanext-scheming you are able to add a field to your schema but have it hidden in the web form UI by using `"form_snippet": null`.  However, when a user edits the package the value/key are removed from the package object.

This validator allows you to hide the `form_snippet` without losing the value of the field.  This does not create a read only version as in this use-case, the desire is to just not lose the existing value for "hidden" form fields. A read only value may require its own validator to prevent editing if a value is provided.  This validator allows editing of the value via an API call for instance.

I've suggested this for scheming as the hidden form aspect that it's tied to is here instead CKAN core.

This was more or less lifted from ckanext-canada.

Related PRs / Issues:
PR #168 
Issue #66 
https://github.com/ckan/ckan/issues/1839#issuecomment-48755345

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -40,6 +40,7 @@ from ckanext.scheming.validation import (
     scheming_isodatetime_tz,
     scheming_valid_json_object,
     scheming_load_json,
+    scheming_do_not_change_if_missing,
 )
 from ckanext.scheming.logic import (
     scheming_dataset_schema_list,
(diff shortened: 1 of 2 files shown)


wardi:
this won't work for resource fields, group/org fields or fields that are part of the original package schema. I think this feature would be better implemented in ckan core package_update by adding the original dict to the context passed through to validators.

boykoc:
Thanks for the feedback. True on not working for any other non-package fields.

I'm going to leave here just until I get a working version for ckan core as suggested.

PR #248, opened 2020-06-22
add display snippet for file size

RabiaSajjad:


wardi:
we can add this snippet to ckanext-canada instead because it's not being used by anything in this repo

wardi:
fixed by #412

PR #249, opened 2020-06-23
Restore auto-generation of resource name

smotornyuk:
I think, starting from the latest patch-releases, the resource name won't be generated using the name of the uploaded file. 

There was [name attribute](https://github.com/ckan/ckan/blob/2.7/ckan/templates/package/snippets/resource_form.html#L24) added to [image_upload macro](https://github.com/ckan/ckan/blob/2.7/ckan/templates/macros/form.html#L409). The value of this attribute is used for selecting input(by the `name` HTML attr) which will be filled with the auto-generated name. The current version of presets doesn't have such value and default version from macro(`image_url`) won't work, because, in case of resource form, name of the field is `name`

### DIFF
--- a/ckanext/scheming/presets.json
+++ b/ckanext/scheming/presets.json
@@ -47,7 +47,8 @@
         "form_placeholder": "http://example.com/my-data.csv",
         "upload_field": "upload",
         "upload_clear": "clear_upload",
-        "upload_label": "File"
+        "upload_label": "File",
+        "upload_name": "name"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


wardi:
@smotornyuk I had to revert this because it caused an error on a 2.6 site, we need to either find a workaround or drop support for 2.6 in ckanext-scheming

smotornyuk:
I don't think that such a small change worth dropping support of 2.6. I'll check, how it can be done in a more general way, not braking anything

Issue #251, opened 2020-07-18
On CKAN 2.9 when enabling scheming_organizations plugin, New and Edit forms return 500


Engerrs:
Enabling **scheming_organizations** plugins on CKAN 2.9 returns 500 error. My suggestion is that the issue is [here](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/plugins.py#L359), because no "**self**" is added, as a result, when CKAN tries to call the method [here](https://github.com/ckan/ckan/blob/dev-v2.9/ckan/views/group.py#L50), it sends two arguments with **self** and **group_type**, then throws error "TypeError: group_form() takes from 0 to 1 positional arguments but 2 were given".
The solution is to add the missing self arguments to this method in **SchemingOrganizationsPlugin** class.

PR #252, opened 2020-07-18
[#251] Fix New and Edit forms load

Engerrs:


smotornyuk:
There is[ exactly the same problem with groups](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/plugins.py#L329) - could you fix it as well?

PR #253, opened 2020-08-13
Composite remove subfields individually 

maxFredenburgh:
1. Remove subfields individually
2. Add pop-up confirmation window to confirm removal
3. Make previous subfields' displays less prominent

### DIFF
--- a/ckanext/scheming/fanstatic/scheming-repeating-subfields.js
+++ b/ckanext/scheming/fanstatic/scheming-repeating-subfields.js
@@ -6,7 +6,7 @@ this.ckan.module('scheming-repeating-subfields', function (jQuery, _) {
                 $template = $this.children('div[name="repeating-template"]'),
                 template = $template.html(),
                 $add = $this.children('a[name="repeating-add"]'),
-                $remove = $this.children('a[name="repeating-remove"]');
+                $remove = $this.find('a[name="repeating-remove"]');
             $template.remove();
 
             $add.on('click', function(e) {
@@ -23,10 +23,17 @@ this.ckan.module('scheming-repeating-subfields', function (jQuery, _) {
 
             $remove.on('click', function(e) {
                 var $groups = $this.find('.scheming-subfield-group'),
-                    $last = $groups.last(),
+                    $curr = $(this).closest('.scheming-subfield-group'),
+                    index = $curr.data('groupIndex'),
+                    field = $curr.data('field'),
                     count = $groups.length;
                 if(count !== 1) {
-                    $last.remove();
+                    document.getElementById("statusText").innerHTML = "Are you sure you want to remove "+field+" "+(index+1)+"?";
+                    $("#confirmRemoval").on('click', function(e) {
+                        $curr.remove();
+                    });
+                }else{
+                    document.getElementById("statusText").innerHTML = "Can not remove, at least one "+field+" must be present";
                 }
                 e.preventDefault();
             });
--- a/ckanext/scheming/fanstatic/scheming-repeating-subfields.js
+++ b/ckanext/scheming/fanstatic/scheming-repeating-subfields.js
@@ -28,12 +28,17 @@ this.ckan.module('scheming-repeating-subfields', function (jQuery, _) {
                     field = $curr.data('field'),
                     count = $groups.length;
                 if(count !== 1) {
-                    document.getElementById("statusText").innerHTML = "Are you sure you want to remove "+field+" "+(index+1)+"?";
+                    document.getElementById("statusText1").innerHTML = (index+1)+"?";
+                    document.getElementById("statusText0").style.display = 'inline';
+                    document.getElementById("statusText1").style.display = 'inline';
+                    document.getElementById("statusText2").style.display = 'none';
                     $("#confirmRemoval").on('click', function(e) {
                         $curr.remove();
                     });
                 }else{
-                    document.getElementById("statusText").innerHTML = "Can not remove, at least one "+field+" must be present";
+                    document.getElementById("statusText0").style.display = 'none';
+                    document.getElementById("statusText1").style.display = 'none';
+                    document.getElementById("statusText2").style.display = 'inline';
                 }
                 e.preventDefault();
             });
--- a/ckanext/scheming/fanstatic/scheming-repeating-subfields.js
+++ b/ckanext/scheming/fanstatic/scheming-repeating-subfields.js
@@ -29,19 +29,13 @@ this.ckan.module('scheming-repeating-subfields', function (jQuery, _) {
                 document.getElementById("statusText1").innerHTML = (index+1)+"?";
                 document.getElementById("statusText0").style.display = 'inline';
                 document.getElementById("statusText1").style.display = 'inline';
-                document.getElementById("statusText2").style.display = 'none';
-                //if(count !== 1){
-                    $("#confirmRemoval").on('click', function(e) {
-                        if(count ==1){
-                            $add.click();
-                            count +=1;
-                        }
-                        $curr.remove();
-                    });
-                // }else{
-                //     $add.click();
-                //     $curr.remove();
-                // }
+                $("#confirmRemoval").on('click', function(e) {
+                    if(count ==1){
+                        $add.click();
+                        count +=1;
+                    }
+                    $curr.remove();
+                });
 
                 e.preventDefault();
             });
--- a/ckanext/scheming/fanstatic/scheming-repeating-subfields.js
+++ b/ckanext/scheming/fanstatic/scheming-repeating-subfields.js
@@ -5,8 +5,7 @@ this.ckan.module('scheming-repeating-subfields', function (jQuery, _) {
                 $this = $(this.el),
                 $template = $this.children('div[name="repeating-template"]'),
                 template = $template.html(),
-                $add = $this.children('a[name="repeating-add"]'),
-                $remove = $this.find('a[name="repeating-remove"]');
+                $add = $this.children('a[name="repeating-add"]');
             $template.remove();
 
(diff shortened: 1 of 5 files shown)


ckanext/scheming/fanstatic/scheming-repeating-subfields.js
```diff
@@ -23,10 +23,17 @@ this.ckan.module('scheming-repeating-subfields', function (jQuery, _) {
...
             $remove.on('click', function(e) {
                 var $groups = $this.find('.scheming-subfield-group'),
-                    $last = $groups.last(),
+                    $curr = $(this).closest('.scheming-subfield-group'),
+                    index = $curr.data('groupIndex'),
+                    field = $curr.data('field'),
                     count = $groups.length;
                 if(count !== 1) {
-                    $last.remove();
+                    document.getElementById("statusText").innerHTML = "Are you sure you want to remove "+field+" "+(index+1)+"?";
```
wardi, line 31: Need to make this translatable, look at the other js files in ckan for an example

ckanext/scheming/fanstatic/scheming-repeating-subfields.js
```diff
@@ -23,10 +23,17 @@ this.ckan.module('scheming-repeating-subfields', function (jQuery, _) {
...
+                    $("#confirmRemoval").on('click', function(e) {
+                        $curr.remove();
+                    });
+                }else{
+                    document.getElementById("statusText").innerHTML = "Can not remove, at least one "+field+" must be present";
```
wardi, line 36: is the remove button not part of the repeating element? In what case would this be triggered?

ckanext/scheming/plugins.py
```diff
@@ -331,13 +331,18 @@ def expand_form_composite(data, fieldnames):
...
     for key in sorted(data):
         if '-' not in key:
             continue
         parts = key.split('-')
         if parts[0] not in fieldnames:
             continue
+        if parts[1] not in indexes:
+            indexes.append(parts[1])
         comp = data.setdefault(parts[0], [])
+        # if status index is missing, replace with next status
```
wardi, line 344: this code applies to any composite repeating group of fields, status is just the one we're using in our theme

ckanext/scheming/templates/scheming/snippets/display_field.html
```diff
@@ -5,9 +5,14 @@
...
-        {{ h.scheming_language_text(field.label) }} {{ loop.index }}
-      </header>
+      {% if loop.index == fields|length %}
+        <header class="panel-heading">
+          {{ h.scheming_language_text(field.label) }} {{ loop.index }}
+        </header>
+      {%- else -%}
+        <details>
+          <summary class="panel-heading">{{ h.scheming_language_text(field.label) }} {{ loop.index }}</summary>
+      {%- endif -%}
```
wardi, line 15: this change should go in ckanext-canada, it's a preference for how we display our status field not a sensible default for all composite field users

ckanext/scheming/templates/scheming/snippets/form_field.html
```diff
@@ -58,6 +58,22 @@
 
     <div name="repeating-template" style="display:none">{{ repeating_panel('REPEATING-INDEX0', 'REPEATING-INDEX1') }}</div>
   </fieldset>
+  <div class="modal" id="mid-screen" tabindex="-1" role="dialog" aria-labelledby="mid-screenLabel" aria-hidden="true">
+    <div class="modal-dialog" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title" id="mid-screenLabel">Removal Confirmation</h5>
```
wardi, line 65: this and "Confirm" and "Cancel" below need to be translatable

ckanext/scheming/fanstatic/scheming-repeating-subfields.js
maxFredenburgh: The remove button is repeated with every sub -field. When one is clicked, it finds the closest parent sub-field with `$curr = $curr = $(this).closest('.scheming-subfield-group'),`

ckanext/scheming/plugins.py
maxFredenburgh: I can remove this comment

ckanext/scheming/templates/scheming/snippets/display_field.html
maxFredenburgh: ckanext-canada doesn't seem to have a display_field.html snippet. Would adding one override the current one in scheming?

wardi: yes we load our plugin first so it would override the scheming one. You can do the same with the confirmation dialog if you want to override the bootstrap one with a wet-boew one. `{% ckan_extends %}` and `{% block %}` can be used to override specific sections instead of the whole snippet/template

ckanext/scheming/plugins.py
```diff
@@ -331,13 +331,18 @@ def expand_form_composite(data, fieldnames):
...
+        parts[1] = indexes.index(parts[1])
```
wardi, line 345: `index` is a linear search and this code is inside a loop over every composite field, better to use a dict for O(1) performance. e.g. `indexes[parts[1]] = len(indexes) + 1` instead of `indexes.append(parts[1])` above

ckanext/scheming/fanstatic/scheming-repeating-subfields.js
maxFredenburgh: What is the best way to make it translatable? So far I've used `{% set header = {'en':'Removal Confirmation','fr':'Confirmation de Suppression'} %}` and `{{ h.scheming_language_text(header) }}` for en/fr. Is this general enough?

wardi: no, just use normal gettext calls `_()` so that we can use existing system-wide translations

ckanext/scheming/fanstatic/scheming-repeating-subfields.js
wardi: I see. How about allowing removal of the last one but create a new blank record when that happens?

maxFredenburgh: That's a good idea, I'll add it

maxFredenburgh:
Changes have been made. Text is translatable, added blocks for override in ckanext-canada, remove last subfield replaced with empty subfield, fixed efficiency of indexing.

ckanext/scheming/templates/scheming/snippets/form_field.html
```diff
@@ -58,6 +60,25 @@
...
+            <p id="statusText0">{{ _('Are you sure you want to remove') }} {{ h.scheming_language_text(field.label) }} </p><p id="statusText1"></p>
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-secondary" data-dismiss="modal">{{ _('Cancel') }}</button>
+          <button name="confirmRemoval" id="confirmRemoval" type="button" class="btn btn-primary" data-dismiss="modal">{{ _('Confirm') }}</button>
+        </div>
+      </div>
+    </div>
+  </div>
+{% endblock %}
```
wardi, line 81: The rest of ckan uses confirmation boxes like this: 
https://github.com/ckan/ckan/blob/ckan-2.8.5/ckan/templates/group/snippets/group_form.html#L37
which is much less code to maintain.

Can you see if you can use something like this instead? The text I would use is "Are you sure you want to remove these fields?" because we can't assume it makes sense to stick the field label on the end of a sentence in any language (e.g. it has the wrong capitalization in French). We could allow the user to define a `delete_subfields_confirmation` message in the yaml file to override this text and retrieve it with `h.scheming_language_text`.

maxFredenburgh: Are you referring specifically to delete_button_text block of the link you provided? I can implement that. For the removal message, would something like:

> Status 1

> Are you sure you want to remove this?

be ok?

wardi: I was referring to the `data-module-content="{{ _('Are you sure you want to delete this Group?') }}` example, making it possible to change the confirmation string displayed from the schema.

I like your suggestion but it's not a field it's a group of fields, right? "Group" is overloaded so what do you think of 
> Status 1
> Are you sure you want to remove these values?"

maxFredenburgh:
@wardi I have made the changes with a new module for removing subfields

ckanext/scheming/templates/scheming/snippets/form_field.html
```diff
@@ -58,6 +60,25 @@
...
+
+{% block removal_popup %}
+  <div class="modal" id="mid-screen" tabindex="-1" role="dialog" aria-labelledby="mid-screenLabel" aria-hidden="true">
+    <div class="modal-dialog" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title" id="mid-screenLabel">{{ _('Confirmation of Removal') }}</h5>
+        </div>
+        <div class="modal-body" style="text-align:center;">
+            <p id="statusText0">{{ _('Are you sure you want to remove') }} {{ h.scheming_language_text(field.label) }} </p><p id="statusText1"></p>
```
wardi: It's better to include the full sentence with a placeholder so that it can be properly formatted in different languages. Can you try this?
```suggestion
            <p id="statusText0">{{ _('Are you sure you want to remove {0}?').format(h.scheming_language_text(field.label) }} </p><p id="statusText1"></p>
```

maxFredenburgh: @wardi this code is outdated, does not apply in the latest commit

PR #255, opened 2020-09-03
Fix unflattening extras in groups

Zharktas:
Ported a fix from https://github.com/ckan/ckan/pull/5546 for extras not in schema ending up broken in group/organization_show.

Note: this only affects new patch releases of ckan eg. 2.8.5 and 2.9.0


Zharktas:
https://github.com/ckan/ckan/pull/5611 fixes it it correctly and this is unneeded.

PR #260, opened 2020-10-05
replace pop-up with simple removal message

maxFredenburgh:


ckanext/scheming/fanstatic/scheming-repeating-subfields.js
```diff
@@ -17,8 +25,22 @@ this.ckan.module('scheming-repeating-subfields', function (jQuery, _) {
...
+                $count+=1;
                 e.preventDefault();
             });
+
+            $remove.on('click', function(e) {
+                var $groups = $this.find('.scheming-subfield-group'),
+                    $curr = $(this).closest('.scheming-subfield-group'),
+                    $body = $curr.find('.panel-body'),
+                    index = $curr.data('groupIndex'),
+                    field = $curr.data('field');
```
wardi, line 37: what are the index and field variables for?

ckanext/scheming/templates/scheming/snippets/form_field.html
```diff
@@ -53,8 +35,12 @@
   {# A complex field with sub-fields, which can optionally be repeatable. #}
   {% set flat = h.scheming_flatten_subfield(field, data) %}
   {% set flaterr = h.scheming_flatten_subfield(field, errors) %}
-  <fieldset name="scheming-repeating-subfields" class="scheming-fieldset" data-module="scheming-repeating-subfields">
-    {% set alert_warning = h.scheming_language_text(field.form_alert_warning) %}
+  {% block removal_text %}
+    {% set locale = h.dump_json({'content': _('These fields have been removed, click save below to update the record with your changes.') }) %}
+
+<fieldset name="scheming-repeating-subfields" class="scheming-fieldset" data-module="scheming-repeating-subfields" data-module-i18n="{{ locale }}">
+  {% endblock %}
```
wardi, line 42: little weird to end a block at a different html tag level, is this just to allow the locale variable to be visible?

maxFredenburgh: Yes, I just didn't want to include the entire fieldset element but it doesn't make much of a difference where it goes. Should I include the entire element or just the locale variable?

ckanext/scheming/fanstatic/scheming-repeating-subfields.js
maxFredenburgh: I kept them in incase they are needed for the flash message as well

wardi: the flash messages would be issued in the controller, not from JS

ckanext/scheming/templates/scheming/snippets/form_field.html
wardi: I see there's an "options" section in the module definition and the string is repeated in the `set` statement here. What is the purpose of the options section? Apologies for my limited JS knowledge, if it's easier to explain on a call we could do that instead.

maxFredenburgh: The string in the module options is the default text, which can be changed through the locale variable. The locale var is included in the block to allow for a custom message and in this case is just a place holder

ckanext/scheming/templates/scheming/snippets/form_field.html
```diff
@@ -35,11 +37,9 @@
   {# A complex field with sub-fields, which can optionally be repeatable. #}
   {% set flat = h.scheming_flatten_subfield(field, data) %}
   {% set flaterr = h.scheming_flatten_subfield(field, errors) %}
-  {% block removal_text %}
```
maxFredenburgh: @wardi removed block and changed name of content to removal-text

maxFredenburgh:
please ignore my embarrassing placeholders that I forgot to take out before committing

ckanext/scheming/templates/scheming/snippets/form_field.html
```diff
@@ -5,7 +5,9 @@
     <div class="panel panel-primary">
       <header class="panel-heading">
         {{ h.scheming_language_text(field.label) }} {{ index1 }}
+        {{ h.are_there_flash_messages() }}
       </header>
+      <p>MEME</p>
```
wardi: ?

maxFredenburgh: ¯\_(ツ)_/¯

maxFredenburgh: Used it as a placeholder and forgot to remove it

PR #269, opened 2020-11-29
Fix #268

### LINKED ISSUE #268: multiple_select selects extra elements
If a multiple choices allows both "world" and "hello world" as values, and the multiple choice contains "hello world", also "hello" is displayed.

https://github.com/ckan/ckanext-scheming/blob/367d50f00492de757a6acec30d3641e437ba2b74/ckanext/scheming/templates/scheming/display_snippets/multiple_choice.html#L5

Possible fix:

```
{%- set values = data[field.field_name].split(',') -%}
{%- set labels = [] -%}

{%- for choice in h.scheming_field_choices(field) -%}
    {%- for value in values -%}
        {%- if choice.value == value -%}
          {%- do labels.append(h.scheming_language_text(choice.label)) -%}
        {%- endif -%}
    {%- endfor -%}
{%- endfor -%}
```

frafra:
Avoid showing choices that have not been selected.


wardi:
@frafra would you take a look at the test failures, and if possible add a test to cover the case that you're fixing

frafra:
Sure; it seems that the tests send a list, not a string, so there is no need to split it. It could be that I am getting different results as I use different presets for multiple_choices with auto completion.

wardi:
are you using the `scheming_multiple_choice_output` output validator on the field? The presets apply it automatically

frafra:
Ok, I think I pinpointed the issue.
`scheming_multiple_choice_output` is the correct validator to put in place, but there is a problem when loading the data again.
```
{{"selected " if val in data[field.field_name] }}
```
This should probably look like this:
```
{{"selected " if val in data.get(field.field_name, '').split(',') 
```

frafra:
`scheming_multiple_choice_output` is indeed a good suggestion, but mine problem was probably due to `multiple_select.html`. I created a new PR for that: https://github.com/ckan/ckanext-scheming/pull/325

Issue #284, opened 2021-03-23
multiple multiple_text fields additional values are assigned to first instance of multiple_text field


Conzar:
If there are more than 1 multiple_text fields in a schema, additional multiple_text field values (> 1) will be assigned to the very first instance of the multiple_text field.  The first value of each additional multiple_text field are correct.  It is only when there are 2, 3, extra additional values that the problem exists.

### Example Schema

```yml
- field_name : 'multiple_text_1'
    label      : 'Multiple Text 1'
    preset     : multiple_text

  - field_name : 'multiple_text_2'
    label      : 'Multiple Text 2'
    preset     : multiple_text
```

### Example Editing

![image](https://user-images.githubusercontent.com/2646975/112088593-fd20d300-8bf4-11eb-8d7d-c34d289668f6.png)

### Example Display after edit

![image](https://user-images.githubusercontent.com/2646975/112088649-175ab100-8bf5-11eb-9c91-c88f156d8e82.png)

### Example edit after viewing the display

![image](https://user-images.githubusercontent.com/2646975/112088702-280b2700-8bf5-11eb-8846-ae7b0f186d15.png)

### JSON from action api

```json
multiple_text_1":[
         "Text 1 Text A",
         "Text 1 Text B",
         "Text 2 Text B"
      ],
      "multiple_text_2":[
         "Text 2 Text A"
      ],
```

### Summary

I have discussed on the ckan gitter with another user.  Unfortunately, neither of us know enough about how the repeating and javascript code works to provide a patch.  It was suggested to look at https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/assets/js/scheming-multiple-text.js

ccancellieri:
https://github.com/ckan/ckanext-scheming/pull/285

ccancellieri:
Looks done, not sure if someone want's to evaluate it for a merge.
I'm waiting for this merge to propose our DCAT scheming yaml file we need this fix (or a better one) to make it working with the official plugin. ;)

Conzar:
I tested your branch and it works!!!  See screenshots below:

![image](https://user-images.githubusercontent.com/2646975/112923989-ceae7500-916b-11eb-92b1-1295a22a2260.png)
![image](https://user-images.githubusercontent.com/2646975/112924025-dc63fa80-916b-11eb-838d-280c7d080d93.png)

Conzar:
PR was merged into master branch.  Thank you for all your efforts!

PR #285, opened 2021-03-23
#284 Now the add buttow will point to the right component even if we have …

ccancellieri:
I've ported the changes we made (on our customization) but I'm not able to test.
Also I see that it's a bit intrusive so feel free to evaluate this pull req only as an Idea.

### DIFF
--- a/ckanext/scheming/assets/js/scheming-multiple-text.js
+++ b/ckanext/scheming/assets/js/scheming-multiple-text.js
@@ -14,7 +14,7 @@ this.ckan.module('scheming-multiple-text', function($, _) {
     },
     initialize: function() {
       if (!scheming_multiple_text_init_done) {
-        $(document).on('click', 'a[name="multiple-remove-custom"]', function(e) {
+        $(document).on('click', 'a[name="multiple-remove"]', function(e) {
           var list = $(this).closest('ol').find('li');
           if (list.length != 1){
               var $curr = $(this).closest('.multiple-text-field');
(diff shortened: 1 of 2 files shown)


Conzar:
Greetings.  Thank you for this PR!  I am testing it on my local development.  Indeed, when adding fields, it works. However, the remove button no longer functions :(

ccancellieri:
Sorry I forgot the last part, I also removed the hidden field.
Having an empty hidden field will push in the metadata an empty field when you click on save from the interface.
Actually this may resolve an alignment problem with rest uploaded metadata using api.

ccancellieri:
> Greetings. Thank you for this PR! I am testing it on my local development. Indeed, when adding fields, it works. However, the remove button no longer functions :(

Thanks for testing, apologize for the missing piece (I'm manually porting changes, from our customized branch) hope this will address the task not causing side effects.

Conzar:
Hello.  Thanks again.  I am still having an issue with the 'remove' button.

I did the following to apply the changes:

```bash
wget https://github.com/ckan/ckanext-scheming/pull/285.patch
git am 285.patch
```
The output was the following:
```bash
Applying: Now the add buttow will point to the right component even if we have multiple instances of multiple_text
Applying: missing remove field fix
```

Patrick1Rhode:
Hi @ccancellieri ,

Looks interesting let me try to use it on my project.

ccancellieri:
Thanks @Patrick1Rhode for your assistance, we tested and there were another mistake coming from the (-custom)ized js.
Now looks better @Patrick1Rhode will confirm after tests.

Patrick1Rhode:
Good day @ccancellieri, 

I switched to your branch and everything works fine. Thank you!. 
test

Conzar:
I have tested and both adding and removal buttons work.  The repeated fields get saved to the correct parent.  Thank you!

PR #287, opened 2021-04-16
#219 fix Tag appearance

ccancellieri:
This should address:
https://github.com/ckan/ckanext-scheming/issues/219

Since Select2 comes from the core, I'm wondering if we would like to port the .select2-container class fix to the core instead of patching it here...

Note that I'm intentionally not abusing !important usage, it's not necessary

### DIFF
--- a/ckanext/scheming/assets/styles/scheming.css
+++ b/ckanext/scheming/assets/styles/scheming.css
@@ -34,7 +34,7 @@ a.btn.btn-multiple-remove {
 
 /* Fix Tag fields:
  https://github.com/ckan/ckanext-scheming/issues/219 */
-.select2-container.select2-container-multi.form-control{
-  height:auto;
+.select2-container.select2-container-multi.form-control {
+  height: auto;
   padding: unset;
 }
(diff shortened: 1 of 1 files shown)


wardi:
looks like this css change would affect all the forms on the site, can you target it to the specific fields that need the fix with class selectors?

ccancellieri:
Yep, this should look cleaner, please let me know.

wardi:
@Patrick1Rhode does this change fix the issue for you? It's still not specific to the scheming forms but it is at least more targeted.

ccancellieri:
@wardi thanks for reviewing, actually I didn't find any class related to the scheming fields in the form, Am I wrong?
Do we want to consider adding a marker class for the scheming generated items?
That would help maybe in some cases but it will also affect the code...

Patrick1Rhode:
Hi @wardi, sorry for taking long to respond to this; we did the test with new changes, and I can confirm that everything works fine.

smotornyuk:
Fixed in #301 without CSS styles. @wardi , this one can be closed

PR #290, opened 2021-06-28
Add div around help text to fix formatting

kdwarn:
Without inheriting from the form-group class, the help_text text is not
formatted properly, so add a div around it to fix.

### DIFF
--- a/ckanext/scheming/templates/scheming/form_snippets/upload.html
+++ b/ckanext/scheming/templates/scheming/form_snippets/upload.html
@@ -16,4 +16,6 @@
     )
 }}
 {# image_upload macro doesn't support call #}
-{%- snippet 'scheming/form_snippets/help_text.html', field=field -%}
+<div class="form-group">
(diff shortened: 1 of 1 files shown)


smotornyuk:
Confirmed. 
Before this PR:
image

After this PR:
image

But it still has a visual issue. This space between buttons and help text doesn't look right. I think that it either must be fixed together with the formatting or formatting must be ignored. If the end-user has to fix margins, he can fix formatting as well.

kdwarn:
Moved the div to a different template (not for display purposes, but because more appropriate I think) and added negative margin. How's that look?

kdwarn:
I can also submit a fresh pull request to avoid the multiple commits.

smotornyuk:
You can't hardcode styles for the `help_text.html` because it's used all over the form, but this extra margin is special only for the resource upload field. Here is an example of new issue introduced by such solution:
image

kdwarn:
Ah, correct the first time. Back to that, with negative margin.

PR #295, opened 2021-08-26
Config setting to add custom composite separator

### LINKED ISSUE #294: Default composite separator conflicts with spatial extension
While working with scheming, spatial, harvester, composite, and repeating extensions I noticed that the default separator of '-' for composite fields caused some problems. This was mostly due to the default field names in the spatial extension containing the same character. I had previously forked the composite extension and added a config setting to set the separator used on composite fields to a custom character.

Now that scheming supports this functionality directly I suggest adding a config setting to allow users to set the separator character to something other than a dash if needed.  Pull request https://github.com/ckan/ckanext-scheming/pull/293 addresses this suggestion.

fostermh:
While working with scheming, spatial, harvester, composite, and repeating extensions I noticed that the default separator of '-' for composite fields caused some problems. This was mostly due to the default field names in the spatial extension containing the same character. I had previously forked the composite extension and added a config setting to set the separator used on composite fields to a custom character.

Now that scheming supports this functionality directly I suggest adding a config setting to allow users to set the separator character to something other than a dash if needed. This pull request addresses this suggestion.

To use a custom separator character, add the `scheming.composite.separator` setting to your ckan.ini file.

Fix #294

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ scheming.organization_schemas = ckanext.scheming:org_with_dept_id.json
 #
 # scheming.dataset_schemas = http://example.com/spatialx_schema.yaml
 
+#   The separator to use to flatten composite fields (repeating_subfields)
+scheming.composite.separator = |
+
 #   Preset files may be included as well. The default preset setting is:
 scheming.presets = ckanext.scheming:presets.json
 
--- a/ckanext/scheming/assets/js/scheming-multiple-text.js
+++ b/ckanext/scheming/assets/js/scheming-multiple-text.js
@@ -3,7 +3,7 @@ this.ckan.module('scheming-multiple-text', function($, _) {
   MultipleText = {
 
      multiple_add: function(field_name){
-      var fieldset = $('fieldset[name='+field_name+']');
+      var fieldset = $('fieldset[name=\"'+field_name+'\"]');
       let list = fieldset.find('ol')
       let items = list.find('li')
       var copy = items.last().clone();


ckanext/scheming/helpers.py
```diff
@@ -27,6 +27,9 @@ def lang():
     from ckantoolkit import h
     return h.lang()
 
+@helper
+def scheming_composite_separator():
+    return config.get('scheming.composite.separator','|')
```
ccancellieri, line 32: Probably it would be better to keep this setting as the default '-' and override if needed.
Could this affect existing schemas? If so we should mode this to the scheming yaml / json instead providing a per type setting.

ccancellieri:
I'm testing it with the spatial plugin, and apart few changes (check comments in pull req), it looks working.

ckanext/scheming/helpers.py
fostermh: good point about the default separator character. I have made that change. I don't think this will affect existing schemas, now that it is using the correct default separator char. A per field override is an interesting idea though. It would be a bit of work as everywhere the separator is used we would also need to check which field we are handling and check the schema. So far, my experience is that a global setting is meeting all our needs in cioos.

ccancellieri:
Hi so in short it's a very good Idea,
 I'm testing on my iso19115 and looks a good improvement.
 
 I've still a problem on subfields of type multifield.

That button was not working due to an issue with the new character on the js.

image

I'll send a pull req to you so you can integrate in this pull req.

ccancellieri:
here is it, just a supereasy fix to the js and a line in the documentation:

https://github.com/cioos-siooc/ckanext-scheming/pull/2

fostermh:
ok, I merged in your pull request. thanks for adding the README update. It seems the 2.9 tests are failing but I can't seem to get the tests to run locally. not sure what the issue is there.

ccancellieri:
I'm checking the error and it looks unrelated [#310], although I'm aligning the tests to better reflect the usage of the separator.

This patch looks good (tested, worked) and it's very useful when spatial is involved so thank you!

fostermh:
ok great. I think this is good to go then. Not sure who should review next and/or merge?

PR #299, opened 2021-09-13
replace usage of unicode with unicode_safe

### LINKED ISSUE #298: Replace usage of `unicode` in schemas and presets
This only affects current CKAN master (future 2.10) but flagging in advance

After https://github.com/ckan/ckan/pull/6048 was merged, trying to use a schema that uses builtin functions like `unicode` (py2) or `str` fails with:

```
E           TypeError: str cannot be used as validator because it is not a user-defined function
```

The example schema that this extension ships with, and some of the presets use some of these builtin validators/converters so we should replace them with the alternatives currently used in CKAN core (ie `unicode_safe`).

For previous versions of CKAN I think ckanext-scheming should provide its own copies of the validators, or include them in ckantoolkit.

TomeCirun:
Fix #298

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,13 +413,13 @@ passing the comma-separated values within as string parameters
 and the result is used as the validator/converter.
 
 ```yaml
-validators: if_empty_same_as(name) unicode
+validators: if_empty_same_as(name) unicode_safe
 ```
 
 is the same as a plugin using the validators:
 
 ```python
-[get_validator('if_empty_same_as')("name"), unicode]
+[get_validator('if_empty_same_as')("name"), unicode_safe]
 ```
 
 This string does not contain arbitrary python code to be executed,
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ validators: if_empty_same_as(name) unicode_safe
 is the same as a plugin using the validators:
 
 ```python
-[get_validator('if_empty_same_as')("name"), unicode_safe]
+[get_validator('if_empty_same_as')("name"), get_validator('unicode_safe')]
 ```
 
 This string does not contain arbitrary python code to be executed,
--- a/ckanext/scheming/validation.py
+++ b/ckanext/scheming/validation.py
@@ -8,6 +8,7 @@ import six
 
 import ckan.lib.helpers as h
 from ckan.lib.navl.dictization_functions import convert
+from ckan.lib.navl.validators import unicode_safe
 from ckantoolkit import (
     get_validator,
     UnknownValidator,
@@ -431,7 +432,7 @@ def get_validator_or_converter(name):
     Get a validator or converter by name
     """
     if name == 'unicode':
-        return six.text_type
+        return unicode_safe
     try:
         v = get_validator(name)
         return v
--- a/ckanext/scheming/validation.py
+++ b/ckanext/scheming/validation.py
@@ -409,8 +409,8 @@ def validators_from_string(s, field, schema):
     """
     convert a schema validators string to a list of validators
 
-    e.g. "if_empty_same_as(name) unicode" becomes:
-    [if_empty_same_as("name"), unicode]
+    e.g. "if_empty_same_as(name) unicode_safe" becomes:
+    [if_empty_same_as("name"), unicode_safe]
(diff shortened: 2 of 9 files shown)


TomeCirun:
hey @wardi what's your opinion over this? 
Thanks

smotornyuk:
There is a number of places that can be changed as well(even though they do not affect the logic and can be ignored):

* [this can be dropped or replaced with unicode_safe for backward compatibility](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/validation.py#L433-L434)
* [this test](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/tests/test_helpers.py#L57)
* [yaml definition of the dataset](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/ckan_dataset.yaml#L46)
* [README reference](https://github.com/ckan/ckanext-scheming/blob/master/README.md#validators)

TomeCirun:
@smotornyuk  thanks for your suggestion

TomeCirun:
hey @smotornyuk  does my latest commit seems ok ? 
Thanks

smotornyuk:
> or replaced with unicode_safe for backward compatibility

If you just drop it, it's ok for new schemas. But every person who is using `unicode` validator will get an error. Actually, I like it, because in that way we can be sure that people will replace `unicode` with `unicode_safe`.

But, if you want to hide this change (I'll explain, why you may choose this option) and to make your PR compatible with existing schemas(i.e, `unicode` in schema still can be used), you can write something like 
```
if name == 'unicode':
        name = 'unicode_safe'
```

> does my latest commit seems ok ?

Except for backward compatibility - yes. But you've introduced one issue. the `unicode_safe` validator was added in CKAN v2.8, while this repo guarantees support for CKAN starting from v2.6.
In v2.6 and v2.7 you still have to use `unicode` validator. This means you have to replace this: 
```
    if name == 'unicode':
        return six.text_type
```
with something like
```
    if name == 'unicode' and not ckantoolkit.check_ckan_version('2.8'):
        return six.text_type
```

wardi:
I'd be happy to add a copy of `unicode_safe` to ckanext-scheming so it could be used on 2.7. I'm indifferent about 2.6 as that's not an officially supported release anymore

ckanext/scheming/validation.py
```diff
@@ -557,3 +556,42 @@ def repeating_text_output(value):
         return json.loads(value)
     except ValueError:
         return [value]
+
+
+@scheming_validator
```
wardi, line 561: this is only for validators that are passed the schema information

TomeCirun: Thanks

README.md
```diff
@@ -413,13 +413,13 @@ passing the comma-separated values within as string parameters
...
 ```yaml
-validators: if_empty_same_as(name) unicode
+validators: if_empty_same_as(name) unicode_safe
 ```
 
 is the same as a plugin using the validators:
 
 ```python
-[get_validator('if_empty_same_as')("name"), unicode]
+[get_validator('if_empty_same_as')("name"), unicode_safe]
```
wardi, line 422:
```suggestion
[get_validator('if_empty_same_as')("name"), get_validator('unicode_safe')]
```

ChristianF88:
Hi everyone. May I ask, what's gonna happen with this? I am trying to migrate to 2.10 and am experiencing some trouble with this. I guess once more ppl migrate, more will be affected? There are no more logs for the failed tests so I am not sure what the issue is. If it's something I can help fix, I'm up for it!

wardi:
I don't see a way to re-run tests from the web UI, @TomeCirun if you want to merge again we should be able to see the reason for the test failures, or @ChristianF88 if you like you could create another PR based on this branch

ChristianF88:
> I don't see a way to re-run tests from the web UI, @TomeCirun if you want to merge again we should be able to see the reason for the test failures, or @ChristianF88 if you like you could create another PR based on this branch

Hi @TomeCirun, I'm not sure how to interpret the thumbs up. Are you gonna merge again? Cheers

PR #308, opened 2021-10-25
Extend existing schemas

smotornyuk:
Allow extending existing schemas.

it can be quite handy when one wants to:

* Use a common set of fields among multiple schemas, when each schema differs from others only by 1-2 fields
* Use some sort of standard schema (IDK, maybe some ISO set of fields, or DCAT schema) and slightly customize particular properties of a single field, while keeping all the other fields up-to-date with the repository that defines them
* Aggregate custom fields from different extensions into a single schema. This allows extensions to define "partial"-schema, like so: 
  ```
  dataset_type: _spatial_dataset
  dataset_fields:
  - name: spatial
    validators: not_empty
  ```
  and then final schema of your portal may look so:
   ```
  dataset_type: my_main_dataset_type
  extensions:
  - name: ckan_dataset
     strategy: append
  - name: _spatial_dataset
     strategy: append

  dataset_fields: []
  resource_fields: []
  ```

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -668,9 +668,9 @@ def _extend_schema(schema, schemas):
 
 
 class ExtensionStrategy(enum.Enum):
-    append = enum.auto()
-    merge = enum.auto()
-    replace = enum.auto()
+    append = 1
+    merge = 2
+    replace = 4
 
     def extend(self, fields, parent_fields):
         result = OrderedDict()


jqnatividad:
Hi @smotornyuk, @wardi ,
Just wondering if this PR is still WIP and being considered for merging.
Thanks!

wardi:
I avoided inherited schemas to keep schema design simple and flat. Instead presets can define fields that are reused and allow individual values to be overridden in an obvious way.

With presets it's simple and obvious how to remove an inherited field or reorder inherited field. The same isn't true for inherited schemas.

Not saying I can't be convinced, but I think designing things to be simple is a really nice goal, and 90% of what you can do with inherited schemas you can already do with presets.

smotornyuk:
Yes, I just sometimes feel lazy when copying similar field configuration or even simple `{preset: x}` definitions :)

wardi:
pro tip: if you put your `field_name` in your preset you don't even need to copy that

Issue #305, opened 2021-10-17
Installation fails on python 2.7


ccancellieri:
Hi I'm installing a fresh copy of this plugin over my docker instance and I'm getting:

```
...
Adding ckanext-scheming 2.1.0 to easy-install.pth file

Installed /srv/app/src_extensions/ckanext-scheming
Processing dependencies for ckanext-scheming==2.1.0
Searching for pyyaml
Reading https://pypi.org/simple/pyyaml/
Downloading https://files.pythonhosted.org/packages/36/2b/61d51a2c4f25ef062ae3f74576b01638bebad5e045f747ff12643df63844/PyYAML-6.0.tar.gz#sha256=68fb519c14306fec9720a2a5b45bc9f0c8d1b9c72adf45c37baedfcd949c35a2
Best match: PyYAML 6.0
Processing PyYAML-6.0.tar.gz
Writing /tmp/easy_install-hdquDQ/PyYAML-6.0/setup.cfg
Running PyYAML-6.0/setup.py -q bdist_egg --dist-dir /tmp/easy_install-hdquDQ/PyYAML-6.0/egg-dist-tmp-lli1fC
Traceback (most recent call last):
  File "/srv/app/src_extensions/ckanext-scheming/setup.py", line 43, in <module>
    """,
  File "/usr/lib/python2.7/site-packages/setuptools/__init__.py", line 162, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/lib/python2.7/site-packages/setuptools/command/develop.py", line 38, in run
    self.install_for_development()
  File "/usr/lib/python2.7/site-packages/setuptools/command/develop.py", line 156, in install_for_development
    self.process_distribution(None, self.dist, not self.no_deps)
  File "/usr/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 758, in process_distribution
    [requirement], self.local_index, self.easy_install
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 782, in resolve
    replace_conflicting=replace_conflicting
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1065, in best_match
[...]

PR #309, opened 2021-10-27
set dependecy of pyyaml to version 5.4

sabinem:


ccancellieri:
can be related to #305

smotornyuk:
I may be wrong, but it shouldn't be possible. `pip` will not try to install a library if it has a minimum required python version specified. i.e, if `pyyaml` > 5.4 has `python_requires > 3.0` flag, pip will never try to install it on python2. If this flag is missing, it's an issue of the dependency(`pyyaml` in this case) and must be solved in corresponding repository

PR #313, opened 2021-11-30
[Documentation] add note about exposed api methods

roc:


README.md
```diff
@@ -471,6 +471,49 @@ Allow HTML inside the help text if set to `true`. Default is `false`.
 
 Display help text inline if set to `true`. Default is `false`.
 
+Action API Endpoints
+====================
+
+The extension adds helper methods which expose any configured schemas via:
```
wardi, line 477: these are action endpoints, *not* helper functions.

There are also helper functions defined in https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/helpers.py

roc: Ah yeah I'm wrongly using those terms interchangeably here. Amend incoming 😅

README.md
```diff
@@ -471,6 +471,51 @@ Allow HTML inside the help text if set to `true`. Default is `false`.
...
+    dataset_type: "dataset",
+    about: "A reimplementation of the default CKAN dataset schema",
+    about_url: "http://github.com/ckan/ckanext-scheming",
+    dataset_fields: [...],
+    resource_fields: [...]
+  }
+}
+```
+
+The full list of helper methods is available in [ckanext/scheming/logic.py](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/logic.py)
```
wardi, line 518: still not helpers :-)

roc: True that, done!

PR #315, opened 2022-01-28
Add form_attrs to repeating fields

smotornyuk:
At the moment `repeating_fields.html` [makes an attempt to set form_attrs via macro.input_block](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/form_snippets/repeating_subfields.html#L53). This macro [doesn't have such option](https://github.com/ckan/ckan/blob/master/ckan/templates/macros/form.html#L285)  and instead of the expected result, all the keys in `form_attrs` appended as extra classes to the rendered block :)

Instead, wrap subfields-fieldset into an extra `div` that contains specified form attributes.


wardi:
thank you @smotornyuk !

PR #316, opened 2022-01-28
Allow data-modules inside repeating subfields

smotornyuk:
At the moment, `[data-module]` functionality works only for subfields that were rendered on page load(because CKAN initializes them).

This PR initializes `data-module` inside every subfield that was added in runtime.

I've updated js a bit in addition

### DIFF
--- a/ckanext/scheming/assets/js/scheming-repeating-subfields.js
+++ b/ckanext/scheming/assets/js/scheming-repeating-subfields.js
@@ -1,29 +1,51 @@
-this.ckan.module('scheming-repeating-subfields', function($, _) {
+ckan.module('scheming-repeating-subfields', function($) {
   return {
     initialize: function() {
-      var container = this;
-      var $this = $(this.el);
-      var $template = $this.children('div[name="repeating-template"]');
-      var template = $template.html();
-      var $add = $this.find('a[name="repeating-add"]');
+      $.proxyAll(this, /_on/);
+
+      var $template = this.el.children('div[name="repeating-template"]');
+      this.template = $template.html();
       $template.remove();
 
-      $add.on('click', function(e) {
-        var $last = $this.find('.scheming-subfield-group').last();
+      this.el.find('a[name="repeating-add"]').on("click", this._onCreateGroup);
+      this.el.on('click', 'a[name="repeating-remove"]', this._onRemoveGroup);
+    },
+
+    /**
+     * Add new group to the fieldset.
+     *
+     * Fields inside every new group must be renamed in order to form correct
+     * structure during validation:
+     *
+     *  ...
+     *  (parent, INDEX-1, child-1),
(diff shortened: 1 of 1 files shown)


wardi:
@smotornyuk I'll defer to your JS expertise on these changes. I'm interested in what sort of form fields you're using that have data modules, though.

smotornyuk:
Right now I'm working on a portal, that replicates to some extent iso19115 metadata-export standard. This standard has a lot of dependencies and the most horrible thing is the fact that depending on the value of some nested field **C** (`a->b->C`), the list of possible values for the other nested field **Z** (`x->y->Z`)  can change. Of course, this is validated on the backend side, but we want to provide more or less convenient UI, thus we have to add some JS.

For the single case, we can just write some global handler, that will track all the changes on the page and react only on relevant ones( [like here](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/assets/js/scheming-repeating-subfields.js#L11) - all clicks inside repeating panel are tracked, but jQuery ignores the ones that came not from **remove** button)

But we have a number of similar issues(we are using a custom autocomplete module, custom date pickers, custom select-boxes, etc) and if we handle all of them via global listeners, it can slow down the app in the client browser. And, in addition, I just prefer to narrow down the scope and area of the handler, so I'm adding a lot of small `data-module`s to every single widget instead of adding one huge `data-module`, that will rule the page.

With this change, if I have some custom date picker in my repeating panel, it will be properly initialized when I add a new item. Without it I had to write an extra code, that will [track the moment when an item was created](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/assets/js/scheming-repeating-subfields.js#L40) and initialize the module.

Even though it's not a difficult thing, it doesn't feel natural. Imagine editing the dataset with a repeating subfield. If this dataset already had two repeating panels, they will be rendered on page load by the server, and data modules will be initialized. But when you add an extra item in the browser, you are getting a slightly different panel, because the dynamically created panel hasn't initialized its module. It's not a problem as it can be fixed in 5 minutes, but I feel like it's unexpected behavior, thus I've changed it:)

I also thought about providing an ability to add repeating fields *inside* repeating fields, in order to make the following fields of the dataset possible:

```
contact:
   - organization1
     name1
     individuals:
       - name: X
         email: Y
       - name: A
         email: B 

   - organization2
     name2
     individuals:
       - name: X
         email: Y
       - name: A
         email: B 
   - ...
```

but I think it may lead to overcomplicated cases. If scheming provides such a feature out of the box, people may think that scheming **recommend** to use complex solutions :face_exhaling:

wardi:
@smotornyuk repeating subfields in repeating subfields is certainly doable, but like you I don't love the idea of really complicated forms so it wasn't a priority. IIRC the repeating form JS code and form error reporting were two things not designed with nesting in the current implementation.

PR #319, opened 2022-03-04
create anchor link for better nav

habibayoub:


wardi:
@habibayoub this changes the look of the default templates which other users aren't going to want. Instead you can create a block wrapping around just the heading that we want to change, then make the actual change in ckanext-canada.

Issue #185, opened 2018-07-16
[Solved] Implement radio preset


kimepe:
Hello guys.
I'm trying to build a custom form with fields displaying radio buttons instead of [select] fields (I want users to be able to only choose one single option among several proposals). Is there a way to do that? I've not been yet able to figure it out.
Thanks.

metaodi:
This should be trivial. Use the "choice" preset und simply change the form_snippet. Currently it uses a "<select>" to display the values, you can create your own form_snippet using radio buttons.

kimepe:
Thanks a lot, @metaodi . Your insight was great. I was able to manage it properly. This is basically what I did, if it can help anyone else out there : 

1. I created in the **~\templates\scheming\form_snippets\** folder a file called **radio.html** (you can call it whatever you want)
in that file, I used as a basis the content of the multiple_checkbox.html file, adjusting it in order to make use of radio buttons as follows : 
`
{%- call form.input_block(
    label=h.scheming_language_text(field.label),
    classes=['control-medium'],
    error=errors[field.field_name],
    is_required=h.scheming_field_required(field)) -%}
    <fieldset class="radios">
        {%- for c in field.choices -%}
            <label for="field-{{ field.field_name }}-{{ c.value }}">
                <input id="field-{{ field.field_name }}-{{ c.value }}"
                    type="radio"
                    name="{{ field.field_name }}"
                    value="{{ c.value }}"
                    {{"checked " if c.value in data[field.field_name] }} />
                {{ h.scheming_language_text(c.label) }}
            </label>
        {%- endfor -%}
    </fieldset>
    {%- snippet 'scheming/form_snippets/help_text.html', field=field -%}
{%- endcall -%}
`
Of course, you can customize it with your CSS to your liking

2. In the **scheming.json** file, all the fields which were supposed to display a radio button had as **preset** the "**select**" option, but in addition to it, I added the previously created **radio.html** as form_snippet
`
        {
            "field_name": "your_field_name_here",
            "label": {
                "en": "your_field_label_here",
                "fr": "your_field_label_here"
            },
            **"preset": "select",**
            **"form_snippet": "radio.html",**
            "choices": [
               {
                    "label": {
                        "en": "1st value here",
                        "fr": "1st value here"
                    },
                    "value": "1st value here"
                },
                {
                    "label": {
                        "en": "2nd value here",
                        "fr": "2nd value here"
                    },
                    "value": "2nd value here"
                },
                {
                    "label": {
[...]

PR #322, opened 2022-03-31
add radio input preset

RabiaSajjad:
adapted from https://github.com/ckan/ckanext-scheming/issues/185


ckanext/scheming/presets.json
```diff
@@ -144,6 +144,14 @@
         "form_snippet": "markdown.html",
         "display_snippet": "markdown.html"
       }
+    },
+    {
+      "preset_name": "radio",
+      "values": {
+        "form_snippet": "radio.html",
+        "display_snippet": "radio.html",
```
wardi, line 152: why not use the existing "select.html" display snippet since both are displayed the same

PR #323, opened 2022-04-04
CKAN 2.10 support

amercader:
@wardi these are changes needed to get tests passing in CKAN 2.10 (IIRC there was an issue with Boostrap 5 and the new composite fields, this PR doesn't address that).
Two things that would be good to get your take on:

1. There were some old requirements pinned in `test-requirements.txt` that don't work with Python 3.9. What do you think about unpinning them for CKAN >= 2.9? These are all CKAN core requirements too, so it's safe to assume they will be already installed (see 7451b3bad889075d6edfa6ff5233dcbafa972577)
2. The schemas included in the extension used the old `unicode` builtin function, which is no longer allowed in CKAN 2.10. Elsewhere we have replaced that by the core `unicode_safe` validator so that's what I did here as well. Alas, `unicode_safe` is not present in [CKAN 2.7](https://github.com/ckan/ckanext-scheming/runs/5817551092?check_suite_focus=true) so what do you would like to do?
  a. Drop support for CKAN 2.7
  b. Copy over the `unicode_safe` function to ckanext-scheming
  c. Copy the `unicode_safe` to ckantoolkit for fallback if using CKAN < 2.8 and use that here
  d. something else

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
       image: openknowledge/ckan-dev:${{ matrix.ckan-version }}
     services:
       solr:
-        image: ckan/ckan-solr-dev:${{ matrix.ckan-version }}
+        image: ckan/ckan-solr:${{ matrix.ckan-version }}
       postgres:
         image: ckan/ckan-postgres-dev:${{ matrix.ckan-version }}
         env:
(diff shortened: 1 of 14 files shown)


wardi:
scheming is already using ckantoolkit so that seems like a good place for `unicode_safe` to live, but I'm also fine with dropping 2.7 support since 2.10 should be out soon. If we do drop 2.7 support I wonder if ckantoolkit could be retired as well?

amercader:
@wardi all green now! :tada:  I added `unicode_safe` to `ckantoolkit` to give a bit more time to users to upgrade

PR #338, opened 2022-08-02
Use a copy of fieldnames parameter in expand_form_composite function call during resource iteration.

cole-fields:
fieldnames object is mutated during iteration over resources and so only the resource in position 0 is correctly processed. Use a copy of dict during iteration. 

Prior to this, resources at position 1:last had the format field-index-subfield in the metadata after updating rather than the expected list of dictionaries.

image

Applicable for ckan 2.9.5 and ckanext-scheming 2.1.0.

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -298,7 +298,7 @@ class SchemingDatasetsPlugin(p.SingletonPlugin, DefaultDatasetForm,
             }
             if resource_composite and 'resources' in data_dict:
                 for res in data_dict['resources']:
-                    expand_form_composite(res, resource_composite)
+                    expand_form_composite(res, resource_composite.copy())
             # convert composite package fields to extras so they are stored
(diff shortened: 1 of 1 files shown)


wardi:
@smotornyuk thank you for offering to take a look at this. @cole-fields would you be able to add a small test that reproduces the issue?

cole-fields:
Sure, there is a jupyter notebook here: https://github.com/cole-fields/repeating-resource-fields/blob/master/repeating-res-subfields.ipynb that demonstrates the issue and the potential fix. I'm not sure if this is what you had in mind, I'm not sure how to formally test it.

To replicate the issue, you would just need a ckan instance with a dataset that has more than 1 resource and at least 1 resource_field that is set to repeating_subfields in the yaml file. Then if you edit a resource other than position 0, the resource field will be in the field-index-subfield format in the metadata, not the list of dicts as expected.

PR #342, opened 2022-08-21
Add CKAN-2.10(csrf) support

TomeCirun:
We recently added `CSRF` protection in `CKAN-2.10`, all `CKAN` extensions that implement the `IBlueprint` interface are exempt, however, `ckanext-scheming` does not use blueprints  (I didn't consider it before)...
In this PR I provide a fix.

### DIFF
--- a/ckanext/scheming/helpers.py
+++ b/ckanext/scheming/helpers.py
@@ -7,7 +7,7 @@ import json
 import six
 
 from jinja2 import Environment
-from ckantoolkit import config, _, check_ckan_version
+from ckantoolkit import config, _
 
 from ckanapi import LocalCKAN, NotFound, NotAuthorized
 
@@ -433,12 +433,3 @@ def scheming_flatten_subfield(subfield, data):
         for k in record:
             flat[prefix + k] = record[k]
     return flat
-
-
-@helper
-def csrf_input():
-    import ckan.plugins as p
-
-    if check_ckan_version(min_version='2.10'):
-        return p.toolkit.csrf_input()
-    return ''
--- a/ckanext/scheming/helpers.py
+++ b/ckanext/scheming/helpers.py
@@ -7,7 +7,7 @@ import json
 import six
 
 from jinja2 import Environment
-from ckantoolkit import config, _
+from ckantoolkit import config, _, check_ckan_version
 
 from ckanapi import LocalCKAN, NotFound, NotAuthorized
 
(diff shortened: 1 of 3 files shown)


TomeCirun:
`AttributeError: module 'ckan.plugins.toolkit' has no attribute 'csrf_input'`
This would work after merging ckan/ckan#7021

ckanext/scheming/templates/scheming/organization/group_form.html
```diff
@@ -34,13 +35,13 @@
             {% block save_text %}
               {%- if action == "edit" -%}
                 {%- if 'humanize_entity_type' in h -%}
-                  {{ h.humanize_entity_type('organization', group_type, 'update label') }}
+                  {{ h.humanize_entity_type('Organization', group_type, 'update label') }}
```
amercader, line 38: I don't think this is needed

ckanext/scheming/helpers.py
```diff
@@ -433,3 +433,12 @@ def scheming_flatten_subfield(subfield, data):
         for k in record:
             flat[prefix + k] = record[k]
     return flat
+
+
+@helper
```
amercader, line 438: The goal on https://github.com/ckan/ckan/issues/7016 was that extensions didn't have to add their own `csrf_input()` helpers. After we add the two mock helpers on the 2.8 and 2.9 branches, `{{ h.csrf_input() }} ` won't fail (see https://github.com/ckan/ckan/pull/7021#pullrequestreview-1081899586).

So this should be merged after the next 2.8 and 2.9 patch releases are out

ckanext/scheming/templates/scheming/organization/group_form.html
TomeCirun: the button is None... because of this check [here](https://github.com/ckan/ckan/blob/34401f000846aa59559c77e755e56656cf1355f9/ckan/lib/helpers.py#L1036)

amercader: Then the problem is that `group_type` doesn't match `organization`. In any case that's a separate issue, so please create a separate PR

ckanext/scheming/helpers.py
TomeCirun: > So this should be merged after the next 2.8 and 2.9 patch releases are out

polite reminder

wardi:
closing in favor of #345

PR #345, opened 2022-10-04
csrf token for group/org forms

wardi:
Add support for CSRF token in overridden group/org forms


wardi:
@TomeCirun does this change work for you?

TomeCirun:
> @TomeCirun does this change work for you?

It works against `MASTER` and  fails against `2.9.6`

PR #348, opened 2022-11-11
Allow adding or overriding classes

avdata99:
Some snippets allow to define `classes` in fields.
This PR add this feature for all form snippets


amercader:
@wardi does that sound reasonable?

PR #351, opened 2022-11-30
Integrate custom extra metadata

pdekraker-epa:
Add a scheming preset for custom metadata key value pairs

Solution previously mentioned in #14 and  #229 has a couple issues:

- All or nothing (is enabled for all schemas)
- Incompatible with multi-page forms (fields appear on each page)
- Cannot control placement within the form

This change adds a preset using the underlying CKAN form element. The associated field must be called "extras" to bypass schema conversion. If there is a better approach please advise.

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -260,6 +260,8 @@ class SchemingDatasetsPlugin(p.SingletonPlugin, DefaultDatasetForm,
         composite_convert_fields = []
         for field_list, destination, is_dataset in fg:
             for f in field_list:
+                if is_dataset and f['field_name'] == 'extras':
+                    continue
                 convert_this = is_dataset and f['field_name'] not in schema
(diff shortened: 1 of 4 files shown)


BonaBeavis:
Can this be merged?

pdekraker-epa:
I believe if functions properly. The added preset is however not documented in the README.

pdekraker-epa:
@wardi could you merge this PR or provide feedback on additional changes that may be required?

wardi:
I don't use this feature because I don't like how users can accidentally cause the site to break by using the same names for their custom extras that are used in the schema. Once we start storing scheming extra fields in `plugin_extras` instead of `extras` this problem will be resolved.

I'm not against merging, though because certainly some sites find this feature useful. I think I would add a warning on the README about users needing to avoid field names used in the schema, at least.

About the name of the feature, what do you think about "free-form extras" or something similar that implies a lack of standardization instead of "custom extras"?

PR #365, opened 2023-03-29
Make org_with_dept_id example match the builtin form

mmeisel:
I was using this example as a reference (as I believe is the intention) when trying to add a single field to organizations without changing anything else. But the actual fields used by ckan (at least in 2.9) are named "description" and "image_url" rather than "notes" and "url". Since these are used by ckan on the organizations page, it's important they remain unchanged. So I think it would be better if the example uses the correct names.

While I was at it, I also made sure everything was using the correct preset (when available), made sure the labels matched, and increased the scheming_version to match the rest of the documentation.

### DIFF
--- a/ckanext/scheming/org_with_dept_id.json
+++ b/ckanext/scheming/org_with_dept_id.json
@@ -1,14 +1,12 @@
 {
-  "scheming_version": 1,
+  "scheming_version": 2,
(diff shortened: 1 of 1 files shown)


wardi:
@mmeisel would you merge master so we can get the tests running on this PR?

PR #372, opened 2023-05-28
Allow literal parameters in validator string

smotornyuk:
Try parsing validator parameters using `ast.literal_eval` before naive splitting by commas.
```yaml
  # old behavior, not changed if cannot parse valid literals
  validators: xxx(hello,world)       # xxx("hello", "world")
  
  # if any argument cannot be parsed, the old strategy is used
  validators: xxx(hello,1)           # xxx("hello", "1")

  # when all arguments are valid, pass them as python objects into the validator
  validators: xxx("hello",1,None)    # xxx("hello", 1, None)

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -553,11 +553,11 @@ Set a `property` attribute on dataset fields displayed as "Additional Info", use
 
 ### `validators`
 
-The `validators` value is a space-separated string of validator and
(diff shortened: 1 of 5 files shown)


tino097 (commented):
Thanks @smotornyuk

PR #373, opened 2023-06-04
[Pydantic]

TomeCirun:


TomeCirun:
@wardi @smotornyuk This is only for reference, it's not fully working yet, there is still a lot to be done.. if you've noticed, i didnt change anything in the schemas.yaml, i made the logic around the existing schema files.

PR #403, opened 2024-01-26
Add the ability to create an arbitrary schema for rendering elsewhere

mutantsan:
Right now I'm working on an extended admin panel and I need to create several configurational pages for different plugins. Since we are using `ckanext-scheming` for all of our CKAN projects, I assume it's a good idea to make it more flexible.

The plugin is oriented toward creating a schema for predefined entity types (dataset, group & organization), but it already has the power to make much more. I suggest allowing defining arbitrary schemas with a unique `schema_id`. This schema could be rendered as independent fields with a `render_fields.html` snippet.

It's defined as a separate extension inside `ckanext-scheming`, that must be enabled to work with (same as other parts of the scheming). The schema follows the **same format** as group & organization schemas, except for the `schema_id` field. That is a unique identifier of a schema. Example of a schema:
```
scheming_version: 2
schema_id: ckanext_notifier
about: An example of a config schema for a fictional extension

fields:
  - field_name: ckanext.ckanext_notifier.enable_notifications
    label: Enable notifications
    validators: default(true) boolean_validator
    preset: select
    required: true
    choices:
      - value: true
        label: Enable
      - value: false
        label: Disable

  - field_name: ckanext.ckanext_notifier.notify_to_email
    label: Notification email
    validators: unicode_safe email_validator
    required: true
    help_text: Specify the email address to which the notification will be sent

  - field_name: ckanext.ckanext_notifier.frequency
    label: Notification frequency in seconds
    validators: default(3600) int_validator
    required: true
    input_type: number
```

When the schema is created and registered via CKAN config, we can fetch it with a `scheming_get_arbitrary_schema` helper. Then in a template, we can render a form like this:
```
<form method="POST">
    {% snippet 'scheming/snippets/render_fields.html', fields=schema.fields, data=data, errors=errors %}

    <button type="submit" class="btn btn-primary">{{ _('Update') }}</button>
</form>
```
And get the next result:
image
_I wrote tests but didn't update the documentation yet. I'm going to do it if we are fine with those changes._

### DIFF
--- a/ckanext/scheming/templates/scheming/package/snippets/package_form.html
+++ b/ckanext/scheming/templates/scheming/package/snippets/package_form.html
@@ -68,30 +68,8 @@
   {%- else -%}
     {%- set fields = schema.dataset_fields -%}
   {%- endif -%}
-  {%- for field in fields -%}
-    {%- if field.form_snippet is not none -%}
-      {%- if field.field_name not in data %}
-        {# Set the field default value before rendering but only if
-        it doesn't already exist in data which would mean the form
-        has been submitted. #}
-        {% if field.default_jinja2 %}
-          {% do data.__setitem__(
-            field.field_name,
-            h.scheming_render_from_string(field.default_jinja2)) %}
-        {% elif field.default %}
-          {% do data.__setitem__(field.field_name, field.default) %}
-        {% endif %}
-      {% endif -%}
-      {%- snippet 'scheming/snippets/form_field.html',
-        field=field,
-        data=data,
-        errors=errors,
-        licenses=c.licenses,
-        entity_type='dataset',
-        object_type=dataset_type
-      -%}
-    {%- endif -%}
-  {%- endfor -%}
+
+  {% snippet 'scheming/snippets/render_fields.html', fields=schema.dataset_fields, data=data, errors=errors, entity_type='dataset', object_type=dataset_type, set_fields_defaults=true %}
 
   {%- if pages -%}
     <input type="hidden" name="_ckan_phase" value="{{ active_page }}" />
--- a/ckanext/scheming/templates/scheming/package/snippets/resource_form.html
+++ b/ckanext/scheming/templates/scheming/package/snippets/resource_form.html
@@ -47,34 +47,7 @@
   {%- endif -%}
 
   {%- set schema = h.scheming_get_dataset_schema(dataset_type) -%}
-  {%- for field in schema.resource_fields -%}
-    {%- if field.form_snippet is not none -%}
-      {%- if field.field_name not in data %}
-        {# Set the field default value before rendering but only if
-        it doesn't already exist in data which would mean the form
-        has been submitted. #}
-        {% if field.default_jinja2 %}
-          {% do data.__setitem__(
-            field.field_name,
-            h.scheming_render_from_string(field.default_jinja2)) %}
-        {% elif field.default %}
-          {% do data.__setitem__(field.field_name, field.default) %}
-        {% endif %}
-      {% endif -%}
-      {# We pass pkg_name as the package_id because that's the only
-      variable available in this snippet #}
-      {%- snippet 'scheming/snippets/form_field.html',
-        field=field,
-        data=data,
-        errors=errors,
-        licenses=c.licenses,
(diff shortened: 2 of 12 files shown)


ckanext/scheming/templates/scheming/package/snippets/package_form.html
```diff
@@ -68,30 +68,8 @@
...
-        data=data,
-        errors=errors,
-        licenses=c.licenses,
-        entity_type='dataset',
-        object_type=dataset_type
-      -%}
-    {%- endif -%}
-  {%- endfor -%}
+
+  {% snippet 'scheming/snippets/render_fields.html', fields=schema.dataset_fields, data=data, errors=errors, entity_type='dataset', object_type=dataset_type, set_fields_defaults=true %}
```
wardi, line 72: looks like licenses isn't passed through here or in the resource_form

wardi:
@mutantsan sounds like a good idea, check the test failures and add some docs for the expected usage

ckanext/scheming/templates/scheming/package/snippets/package_form.html
mutantsan: +

mutantsan:
@wardi I've fixed the tests and wrote some doc.
I'm not sure about the `Tests / CKAN 2.8 (pull_request) Failing after 59s`. It seems, that something went wrong during the initialization of the env for tests. Do you know how to fix it?

ckanext/scheming/templates/scheming/package/snippets/resource_form.html
```diff
@@ -47,34 +47,7 @@
...
-        errors=errors,
-        licenses=c.licenses,
-        entity_type='dataset',
-        object_type=dataset_type,
-        package_id=pkg_name
-      -%}
-    {%- endif -%}
-  {%- endfor -%}
-
+  {% snippet 'scheming/snippets/render_fields.html', fields=schema.resource_fields, data=data, errors=errors, entity_type='dataset', object_type=dataset_type, set_fields_defaults=true %}
```
wardi, line 50: entity_type='resource'?

ckanext/scheming/templates/scheming/snippets/render_fields.html
```diff
@@ -0,0 +1,30 @@
...
+    fields - a list of scheming field dictionaries
+    data - form data fields
+    errors - A dict of errors for the fields
+    entity_type - entity type
+    object_type - object type
+    set_fields_defaults - flag to set the default field values
+{#}
+
+{% for field in fields if field.form_snippet is not none %}
+    {% if field.field_name not in data and set_fields_defaults %}
```
wardi, line 11: All the other templates are indented with 2 spaces. Better to be consistent

setup.py
```diff
@@ -1,6 +1,6 @@
 from setuptools import setup, find_packages
 
-version = '3.0.0'
+version = '3.0.1'
```
wardi, line 3: I'll bump the version number when doing a release

ckanext/scheming/templates/scheming/package/snippets/resource_form.html
mutantsan: I don't know, just copied it from the old template. If you ask me, the code must be refactored to remove  ` licenses=licenses, entity_type=entity_type, object_type=object_type` from here.

ckanext/scheming/templates/scheming/snippets/render_fields.html
mutantsan: Sure, thanks

ckanext/scheming/templates/scheming/snippets/render_fields.html
```diff
@@ -0,0 +1,30 @@
+{#}
+    fields - a list of scheming field dictionaries
+    data - form data fields
+    errors - A dict of errors for the fields
+    entity_type - entity type
+    object_type - object type
+    set_fields_defaults - flag to set the default field values
+{#}
```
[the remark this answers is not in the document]
mutantsan: Fixed

setup.py
mutantsan: Sure

PR #406, opened 2024-04-01
Fix ckan version comparison

alexmorev:


ckanext/scheming/templates/scheming/form_snippets/slug.html
```diff
@@ -1,6 +1,6 @@
 {% import 'macros/form.html' as form %}
 
-{% set read_endpoint = '.read' if h.ckan_version() > '2.9' else '_read' %}
+{% set read_endpoint = '.read' if h.ckan_version().split('.')[1] | int >= 9 else '_read' %}
```
wardi, line 3: A better fix is to use `h.check_ckan_version` here.

alexmorev: Thank you. I've fixed this. Please take a look and merge if it's OK. And is it possible to create a new tag?

PR #412, opened 2024-06-13
Add two new snippets

amercader:
* `number.html` for form fields that require a number value
* `file_size.html` to display resource sizes in a human friendly way


wardi:
is there a validator that should go with the number input? Should the file size display snippet also be used in the example dataset schema?

amercader:
@wardi We have the `int_validator` in core. I added the file size field to the example dataset schema

wardi:
On second thought this adds a field to the form that we probably don't want (also I generally don't like the way number inputs are rendered with a little useless arrow controls). 
number-form
I think we're better without the ckan_dataset.yaml change.

PR #413, opened 2024-06-13
Support for CKAN 2.11

amercader:
Freshen up the worflow file and adapt tests to changes in test client authentication in CKAN 2.11
Starting from CKAN 2.11 test client calls are authenticated via `headers={"Authorization": token}` rather than `environ_overrides={"Authorization": token}`. I added some helper functions to handle different CKAN versions from a single place. Also there is a new form in the page so index-based selectors needed to be updated.

More details on the auth changes here:
https://github.com/ckan/ckan/wiki/CKAN-2.10-to-2.11-migration-tips#authentication-when-using-the-test-client

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
     name: CKAN ${{ matrix.ckan-version }}
     runs-on: ubuntu-latest
     container:
-      image: openknowledge/ckan-dev:${{ matrix.ckan-version }}
+      image: ckan/ckan-dev:${{ matrix.ckan-version }}
     services:
       solr:
         image: ckan/ckan-solr:${{ matrix.ckan-version }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -17,13 +17,13 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        ckan-version: ["2.10", "2.9", "2.9-py2"]
+        ckan-version: ["master", "2.10", "2.9"]
```
amercader, line 20: Using `master` for now as the `ckan/ckan-dev:2.11` images are not yet built

wardi (approved):

.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        ckan-version: ["master", "2.10", "2.9"]
+        ckan-version: ["2.11", "2.10", "2.9"]
```
wardi, line 20: don't want to keep testing against master?

PR #414, opened 2024-06-14
Provide a new plugin for indexing repeating subfields

amercader:
The `scheming_subfields_index` plugin will group the values of the same subfields in a text field that will make the values findable.
They are indexed as `extras_{field_name}__{key}`. `extras_*` is a dynamic `text` Solr field that will allow free-text search on these
values.

Added tests and updated the docs.

@wardi let me know what do you think and if you want to keep the `scheming_nerf_index` plugin or remove it in favour of this one.

Note: the test failure is a Python 2 one, support for which is removed in https://github.com/ckan/ckanext-scheming/pull/413

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,10 +280,13 @@ When using a plain string translations will be provided with gettext:
 This field is the parent of group of repeating subfields. The value is
 a list of fields entered the same way as normal fields.
 
-> **_NOTE:_** CKAN needs an IPackageController plugin with `before_index` to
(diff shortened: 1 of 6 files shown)


wardi (approved):
Looks great, please resolve conflict and merge. We can keep the `scheming_nerf_index` plugin in the code but mark is as deprecated and recommend use of `scheming_subfields_index` instead.

wardi:
Adding a comment as a reminder that it's probably worth dusting this off and updating it for #446

PR #415, opened 2024-06-20
Don't store empty repeating subfields

amercader:
When implementing repeating subfields without any mandatory subfields, if empty values are sent from the form an "empty" item is stored, with all its properties empties:

```
{
    "name": "test-dataset",
    "publisher": [
        {
            "email": "",
            "name": "",
            "type": "",
        }
    ]
}
```

The changes in `expand_form_composite()` fix the issue on the scheming side, preparing a `data_dict` with an empty list for that field:

```
{
    "name": "test-dataset",
    "publisher": []
}
```

Sadly, on the CKAN side the navl functions drop the field entirely (somewhere along [`make_full_schema`](https://github.com/ckan/ckan/blob/0628ead372cb17c96b145af0dcb4cb95b9cb67a8/ckan/lib/navl/dictization_functions.py#L150) and [`get_all_key_combinations`](https://github.com/ckan/ckan/blob/0628ead372cb17c96b145af0dcb4cb95b9cb67a8/ckan/lib/navl/dictization_functions.py#L126)) and it doesn't get stored as an extra, so it's missing from the resulting dataset dict.

```
{
    "name": "test-dataset",
}
```

@wardi Ideally we would want an empty list in the resulting dataset, but do you think a missing field is an improvement?

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -397,15 +397,16 @@ def expand_form_composite(data, fieldnames):
         parts[1] = indexes[parts[1]]
         try:
             try:
-                comp[int(parts[1])]['-'.join(parts[2:])] = data[key]
-                del data[key]
+                if data[key]:
+                    comp[int(parts[1])]['-'.join(parts[2:])] = data[key]
             except IndexError:
                 comp.append({})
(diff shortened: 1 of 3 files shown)


wardi:
It would be nicer to be able to store an empty list, but a missing value is an improvement over a list with one object with blank fields for sure.

PR #422, opened 2024-08-29
SPCXCFZB-1154 / fix xloader issue in ckan-2.10.*

yurijGagarin:
Its a fix for the issue that is related to the certain ckan project


wardi:
fixed by #427

PR #426, opened 2024-10-28
Use Python images in tests

amercader:
Rather than the deprecated Alpine ones, as they cause failures like

https://github.com/ckan/ckanext-scheming/actions/runs/11551377696/job/32148273098?pr=425


wardi (approved):

wardi:
@amercader is this still required for running tests against 2.10 and 2.11?

PR #427, opened 2024-10-30
Check if extras exist before converting

amercader:
In some situations like when validating an already validated dataset we might be missing the `extras` key.

See
https://github.com/ckan/ckanext-dcat/issues/305
https://github.com/ckan/ckan/pull/7571


wardi (commented):
LGTM. All other `output_validators` will need to be idempotent as well, right? Maybe need to document that.

Issue #286, opened 2021-04-16
Default validation (ignore missing) is not uniformally applied 


Patrick1Rhode:
### **The issue comes out when we try to create a package via the API. 
**Some missing validator over certain field types are causing SQL exception (please refer to the cases after the stack).****

**Here is the example of an error based on a field called 'contact_email' which is defined without the ignore missing validator:** 

/usr/lib/ckan/default/bin/python /usr/lib/ckan/default/bin/paster serve --reload /etc/ckan/default/development.ini
Starting subprocess with file monitor
/usr/lib/ckan/default/local/lib/python2.7/site-packages/OpenSSL/crypto.py:12: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in the next release.
  from cryptography import x509
2021-04-16 14:21:55,267 INFO  [ckan.config.environment] Loading static files from public
2021-04-16 14:21:55,375 INFO  [ckan.config.environment] Loading templates from /home/patrick/ckan/lib/default/src/ckan/ckan/templates
2021-04-16 14:21:57,143 INFO  [ckan.config.environment] Loading templates from /home/patrick/ckan/lib/default/src/ckan/ckan/templates
Starting server in PID 14947.
serving on 0.0.0.0:5000 view at http://127.0.0.1:5000
2021-04-16 14:22:03,566 INFO  [ckan.views.api] Validation error (Action API): "{'name': [u'That login name is not available.'], u'__type': u'Validation Error'}"
2021-04-16 14:22:03,814 INFO  [ckan.views.api] Validation error (Action API): "{'name': [u'That login name is not available.'], u'__type': u'Validation Error'}"
2021-04-16 14:22:04,062 INFO  [ckan.views.api] Validation error (Action API): "{'name': [u'That login name is not available.'], u'__type': u'Validation Error'}"
/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py:2181: SAWarning: Usage of the 'related attribute set' operation is not currently supported within the execution stage of the flush process. Results may not be consistent.  Consider using alternative event listeners or connection-level operations instead.
  % method)
[...]

PR #430, opened 2025-01-13
Update presets.json

rausnehmkeil:
Solves issues with "multiple_text" preset.

E.g.: https://github.com/ckan/ckanext-scheming/issues/286


wardi:
I just tested a multiple_text field without any change to the validators and I don't see the error described with the latest version. If this is still happening please provide the steps you used to trigger the error.

PR #431, opened 2025-01-17
is_organization is not set when organization created via UI

smotornyuk:
When creating an organization via WebUI, its `is_organization` flag depends on the value of the plugin's `is_organization` attribute. scheming_organizations doesn't have this attribute and custom organizations are saved as groups.

API uses different logic and is not affected by this issue, so the fix is required only for WebUI.


wardi:
fixed by #437

wardi:
Oh, but I like your test. I'll merge this too

wardi:
merged as part of #438

PR #437, opened 2025-02-26
Fix is_organization for custom organization type schema

avdata99:
CKAN core defines as default `is_organization = False` for the `IGroupForm`
https://github.com/ckan/ckan/blob/master/ckan/plugins/interfaces.py#L1352

Also, CKAN check is a group is an `organization` using this property
https://github.com/ckan/ckan/blob/master/ckan/lib/plugins.py#L223-L226

CKAN group index views at `ckan/views/group.index` uses this `is_organization` property.

If a custom `organization_type` is defined with a schema file through `ckanext-scheming` we need to ensure `is_organization = True`

This PR fixes this.

Also, to release 3.0.1 is proposed. A new tag `release-3.0.1` is required
Also, GitHub action fixed

### DIFF
--- a/ckanext/scheming/plugins.py
+++ b/ckanext/scheming/plugins.py
@@ -460,6 +460,7 @@ class SchemingOrganizationsPlugin(p.SingletonPlugin, _GroupOrganizationMixin,
     FALLBACK_OPTION = 'scheming.organization_fallback'
     SCHEMA_TYPE_FIELD = 'organization_type'
     UNSPECIFIED_GROUP_TYPE = 'organization'
+    is_organization = True
 
     @classmethod
     def _store_instance(cls, self):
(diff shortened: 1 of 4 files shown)


pdelboca:
@avdata99 looks good but I would add a nice verbose comment over that line. 

This touches on some poorly documented behaviors of CKAN so a simple `is_organization = True` without the context of this PR will be difficult to debug and maintain in the future.

ckanext/scheming/plugins.py
```diff
@@ -460,6 +460,7 @@ class SchemingOrganizationsPlugin(p.SingletonPlugin, _GroupOrganizationMixin,
     FALLBACK_OPTION = 'scheming.organization_fallback'
     SCHEMA_TYPE_FIELD = 'organization_type'
     UNSPECIFIED_GROUP_TYPE = 'organization'
+    is_organization = True
```
pdelboca, line 463:
```suggestion
    # IGroupForm sets is_organization = False as a default value. 
    # To properly use it to extend the Organization schema we need to set it to true.
    # See: https://github.com/ckan/ckanext-scheming/pull/437
    is_organization = True
```

avdata99:
This is probably ready to go @pdelboca

pdelboca:
@wardi any feedback on this PR? 

Also, I'm not sure how to handle the release cycle of `ckanext-scheming` but this looks a good candidate for a minor release.

PR #451, opened 2025-11-03
feat: Adding additional templates for Repeating fields

Engerrs:
Out of the box, the template for Repeating fields is straightforward, but the main issue with it is the space it takes, especially where there more the 2 values or fieldset of multiple fields in it.

This PR adds two additional options for displaying Repeating Fields:

1. Accordion where the fields are stored
2. Accordion, but with additional ability to provide grid layout for the fields to save space.

https://github.com/user-attachments/assets/fb78f5d6-da56-4963-8125-88342bcca825

I understand that having 3 different templates is hard to maintain, so I can suggest combaine those into 1 with additional configuration from the schema itself.
Maybe you have other suggestions that I can work on?


wardi:
These would be more maintainable if they were incorporated into the existing `repeating_subfields.html` snippet by adding a couple new field options:
```yaml
repeating_collapse: true
repeating_grid: true
```

Issue #400, opened 2024-01-23
Missing in documentation


boranin:
In my view, answers to the following would enhance the documentation:

-Why is the dataset schema given in .yaml and the group and organisation schemas as JSONs? Can they all be either?
-What is the create_validators attribute used for? Examples of usage would be of help too. 
-field_name - "Existing dataset field names include [...]" - Are these fields defined in the ckan_dataset.yaml file? There are more fields defined in that file than on the [...] list, should they be on the list in the Readme file?
-Can the presets be used for all, datasets, groups and organisation fields or are they specific to datasets only? 

It might be that all this has been answered somewhere, but I didn't see that even after re-reading the Readme several times.

wardi:
@boranin thanks! these are good suggestions for improvements to the docs.

PR #454, opened 2025-11-26
Fix missing documentation #400

mjanez:
Documentation clarifications added to address the questions raised in issue #400: formats, create_validators, field_name usage, presets scope, and repeating labels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,30 @@ for each group.
     label: Phone Number
 ```
 
+* `label`: This is the human-readable label for the **entire group** of repeating fields (e.g., "Contacts"). It typically appears as the header for the section.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -672,8 +702,28 @@ sent to the user.
...
+This is useful for:
+- Fields that are mandatory when creating a dataset but read-only or optional during updates.
+- Setting initial default values or identifiers that shouldn't be validated the same way afterwards.
+
+```yaml
+- field_name: source_id
+  label: Source dataset identifier
+  form_placeholder: 'INT-12345'
+  # Creation: must be present and valid
+  create_validators: scheming_required not_empty unicode_safe
```
wardi, line 717: scheming_required only requires fields when `required: true` is also set.

This example is also a little strange: it would not let someone create a record without a source_id but on update they could choose to remove the value.

PR #459, opened 2026-05-27
CKAN 2.12 support

amercader:
Testing the latest default branch against the ckan/ckan-dev:2.12 image to surface any potential issues

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        ckan-version: ["2.11", "2.10"]
+        ckan-version: ["2.12", "2.11", "2.10"]
       fail-fast: false
 
     name: CKAN ${{ matrix.ckan-version }}
--- a/ckanext/scheming/templates/scheming/package/snippets/resource_form.html
+++ b/ckanext/scheming/templates/scheming/package/snippets/resource_form.html
@@ -40,8 +40,18 @@
 {% block errors %}
   {%- if errors -%}
     {%- set schema = h.scheming_get_dataset_schema(dataset_type) -%}
+    {%- if data -%}
+      {%- set fields = schema.resource_fields %}
+    {%- else -%}
+      {# combined dataset and resource errors #}
+      {%- set fields = schema.dataset_fields + [{
+        'field_name': 'resources', 'label': _('Resources'),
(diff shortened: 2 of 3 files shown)


smotornyuk:
@amercader, the failing test is not related to v2.12. There are a few additional lines in the test that are executed only for v2.12, but they fail because of an incorrect assumption.

@wardi, [here](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/tests/test_form.py#L575-L582) the test expects that on the resource form, invalid dataset fields will be reported with correct labels from the schema. But this is a resource form, and it passes [resource fields](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/package/snippets/resource_form.html#L44), not the dataset fields, into the error snippet. As a result, the snippet cannot produce a correct label for the problematic dataset field.

Additionally, [here ](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/tests/test_form.py#L582), the test expects an error message that is contained inside the `error_summary` dictionary, which is not used in scheming. Instead, scheming generates an error message from `error_dict`(called `errors` in templates).

So, instead of 
```
Name: missing
Description: missing
Resources: Invalid resources
```

You'll see 
```
name: missing
notes: missing
resources: [{name: missing}]
```

I don't see a simple fix for it; the whole error snippet requires potentially backward-incompatible refactoring to manage such a situation. Probably the best option is to pass both the dataset and resource schemas into the error snippet. But another problem is that depending on the field that caused the validation error, the structure of `errors` will be different: 

* if resource's `name` is invalid: `{name: ...}`
* if dataset's `name` is invalid: `{name: ...}`
* if resource's and dataset's `name` fields are invalid: `{name: ..., resources: [{name: ...}]}`

In the last case, the structure becomes more complex, but the real problem is with the first two scenarios. There is no distinction between an error caused by a resource's `name` field and a dataset's `name` field, but we must somehow discover which one caused the problem and render it with a `Name` label if it's a resource, and with a `URL` label if it's a dataset.

wardi:
@smotornyuk I've applied a fix that will display both dataset and resource errors on the same page including the correct labels for both by passing the resource fields as a temporary "repeating_subfields" field in the dataset fields schema.

I'm using the fact that data is passed as `{}` when the resource form renders dataset errors. It's not pretty but seems to work without any core changes required.

PR #460, opened 2026-07-27
Dynamic schemas

wardi:


ckanext/scheming/ckan_formpages_draft.yaml
```diff
@@ -56,7 +56,6 @@ dataset_fields:
 
 - field_name: version
   label: Version
-  validators: ignore_missing unicode_safe package_version_validator
```
wardi, line 59: this seems unrelated

ckanext/scheming_dynamic/logic/validators.py
```diff
@@ -0,0 +1,70 @@
...
+    Only "dataset" entities are backed by an actual model row today (see
+    ``ENTITY_TYPES`` in ``logic/schema.py``), so this is a no-op for any
+    other entity_type.
+    """
+    entity_type = data.get(("entity_type",))
+    if entity_type is tk.missing or not entity_type:
+        entity_type = DEFAULT_ENTITY_TYPE
+
+    if entity_type != DEFAULT_ENTITY_TYPE:
+        return
```
wardi: better to fail with a validation error

ckanext/scheming_dynamic/migration/scheming_dynamic/alembic.ini
```diff
@@ -0,0 +1,74 @@
...
+# version location specification; this defaults
+# to /home/cherry/projects/master/ckanext-scheming/ckanext/scheming_dynamic/migration/scheming_dynamic/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+# version_locations = %(here)s/bar %(here)s/bat /home/cherry/projects/master/ckanext-scheming/ckanext/scheming_dynamic/migration/scheming_dynamic/versions
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+sqlalchemy.url = driver://user:pass@localhost/dbname
```
wardi, line 38: should this be here?

ckanext/scheming_dynamic/schema.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def properties(self) -> dict[str, Any]:
+        return {
+            **super().properties(),
+            "group_type": {"type": "string", "minLength": 1},
+            "fields": self.FIELD_LIST,
+        }
+
+
+class OrganisationSchema(BaseSchema):
```
wardi: American spellings for consistency

```suggestion
class OrganizationSchema(BaseSchema):
```

ckanext/scheming_dynamic/cli.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+schema_type_option = click.option(
+    "-t",
+    "--type",
+    "schema_type",
+    type=click.Choice(sorted(SCHEMA_TYPES)),
+    default="dataset",
+    show_default=True,
+    help="Which ckanext-scheming schema shape to build/validate against.",
```
wardi, line 18: This option should be called *entity type* not schema type or schema shape. Also we should either remove the option or document that only `dataset` is supported.

ckanext/scheming_dynamic/schema.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def properties(self) -> dict[str, Any]:
+        return {
+            **super().properties(),
+            "organization_type": {"type": "string", "minLength": 1},
+            "fields": self.FIELD_LIST,
+        }
+
+
+SCHEMA_TYPES: dict[str, type[BaseSchema]] = {
```
wardi: let's call these either ENTITY_TYPES to match the keys, or METASCHEMA_CLASS to match the values.

ckanext/scheming_dynamic/cli.py
```diff
@@ -0,0 +1,70 @@
...
+
+@click.group(short_help="scheming_dynamic commands")
+def scheming_dynamic():
+    """scheming_dynamic commands"""
+
+
+@scheming_dynamic.command()
+@schema_type_option
+def schema(schema_type: str):
+    """Print the JSON Schema for the given schema type.
```
wardi:
```suggestion
    """
    Output the JSON Schema used to validate ckanext-scheming schemas
    for the given entity type.
```

ckanext/scheming_dynamic/cli.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@click.group(short_help="scheming_dynamic commands")
+def scheming_dynamic():
+    """scheming_dynamic commands"""
+
+
+@scheming_dynamic.command()
+@schema_type_option
+def schema(schema_type: str):
```
wardi: Maybe something like
```suggestion
def schema_validation_json_schema(schema_type: str):
```
would be clearer. Or, we could introduce a new term like "metaschema" for these

ckanext/scheming_dynamic/model.py
```diff
@@ -0,0 +1,67 @@
...
+        sa.Column("entity_type", sa.Text, primary_key=True),
+        sa.Column("schema_type", sa.Text, primary_key=True),
+        sa.Column(
+            "updated",
+            sa.TIMESTAMP(timezone=True),
+            index=True,
+            default=_current_datetime,
+            onupdate=_current_datetime,
+        ),
+        sa.Column("definition", JSONB),
```
wardi:
```suggestion
            nullable=False,
        ),
        sa.Column("definition", JSONB, nullable=False),
```

ckanext/scheming_dynamic/README.md
```diff
@@ -0,0 +1,60 @@
+# scheming_dynamic
+
+Store ckanext-scheming dataset schemas in the database and edit them without
+a server restart.
+
+## How it works
+
+- Schemas live in the `scheming_schema` table (`entity_type` + `schema_type`
+  primary key, JSONB `definition`). `schema_type` is a unique entity type (e.g. `dataset_type`)
+  and must match the `dataset_type` field inside the definition.
```
wardi, line 10: is this check added to the JSON schema for checking the definitions?

ckanext/scheming_dynamic/README.md
```diff
@@ -0,0 +1,60 @@
...
+- Before ckanext-scheming returns any dataset schema it asks this extension
+  whether the table changed since the last merge (a cheap
+  `count(*) + max(updated)` scan, at most once per request). When it did,
```
wardi: With the `updated` index this should be an O(1) lookup, not a scan

ckanext/scheming_dynamic/README.md
```diff
@@ -0,0 +1,60 @@
...
+Definitions are validated against a JSON Schema mirroring the minimal shape
+of a ckanext-scheming schema file; `preset` values are checked against the
+presets registered at startup.
+
+## Known limitations
+
+- Only dataset schemas are merged at runtime; `group`/`organization` rows
+  can be stored but are not yet picked up by ckanext-scheming.
+- Form pages (`start_form_page`) work for types known at startup only: the
+  paged create/edit routes are registered per type when the app boots.
```
wardi: this should be fixable, right? Certainly can be done later though

ckanext/scheming_dynamic/README.md
```diff
@@ -0,0 +1,60 @@
...
+- Changing a schema does not reindex existing datasets; run
+  `ckan search-index rebuild` after incompatible changes.
```
wardi, line 76: also fixable with background jobs later.

ckanext/scheming_dynamic/schema.py
```diff
@@ -0,0 +1,185 @@
...
+    CHOICE = {
+        "type": "object",
+        "required": ["value"],
+        "description": (
+            "`label` is optional: ckanext-scheming's own "
+            "scheming_choices_label() helper falls back to the value "
+            "itself (c.get('label', value)) when a choice has no label."
+        ),
+        "properties": {
+            "value": {"$ref": "#/$defs/anytype"},
```
wardi: IIRC these need to be string values

ckanext/scheming_dynamic/schema.py
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+    def required(self) -> list[str]:
+        return ["about_url"]
+
+    def properties(self) -> dict[str, Any]:
+        return {
+            "about_url": {"type": "string", "format": "uri", "minLength": 1},
+            "about": {"type": "string"},
+            "scheming_version": {"type": "integer"},
```
wardi: this has always been ignored. it could be removed here

ckanext/scheming_dynamic/schema.py
```diff
@@ -0,0 +1,185 @@
...
+            "$defs": self.defs(),
+        }
+
+    def required(self) -> list[str]:
+        return ["about_url"]
+
+    def properties(self) -> dict[str, Any]:
+        return {
+            "about_url": {"type": "string", "format": "uri", "minLength": 1},
+            "about": {"type": "string"},
```
wardi: this doesn't exist, but if it did it would be i18n text

ckanext/scheming_dynamic/schema.py
```diff
@@ -0,0 +1,185 @@
...
+    def defs(self) -> dict[str, Any]:
+        return {
+            "anytype": self.ANYTYPE,
+            "i18n_text": self.I18N_TEXT,
+            "choice": self.CHOICE,
+            "preset": {
+                "type": "string",
+                "description": (
+                    "Must be one of the presets registered at startup: the "
+                    "built-in ckanext/scheming/presets.json plus any files "
```
wardi: `ckanext/scheming/presets.json` needs to be in scheming.presets too

ckanext/scheming_dynamic/schema.py
```diff
@@ -0,0 +1,185 @@
...
+
+        Preset names ckanext-scheming itself would accept: the presets
+        loaded at startup from the scheming.presets config option.
+        """
+        from ckanext.scheming.plugins import _SchemingMixin  # noqa: PLC0415
+
+        if _SchemingMixin._presets is None:
+            _SchemingMixin._load_presets(tk.config)
+
+        return sorted(_SchemingMixin._presets)  # type: ignore
```
wardi: It's worth moving this prefix lazy-loading into _SchemingMixin, in case we need to make reloads thread-safe in the future

ckanext/scheming_dynamic/schema.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class DatasetSchema(BaseSchema):
+    schema_id = "https://github.com/ckan/ckanext-scheming/scheming_dynamic/dataset_schema.schema.json"
```
wardi, line 312: Since the JSON schemas for dynamic scheming schemas are themselves dynamic (dynamic metaschemas?), maybe add a section to the readme and use something like:

https://github.com/ckan/ckanext-scheming/blob/dynamic-schemas/ckanext/scheming_dynamic/README.md#dataset-schema

ckanext/scheming_dynamic/schema_sync.py
```diff
@@ -0,0 +1,105 @@
...
+    check runs at most once per request and once per worker process, which is
+    how changes made in one worker propagate to the others.
+
+    The fingerprint is only advanced once the caller confirms the merged
+    schemas were applied successfully (see ``confirm_applied``) — otherwise a
+    schema that fails to expand would never be retried until some unrelated
+    write bumped the fingerprint again.
+    """
+    if _checked_in_this_request():
+        log.debug("skipping database check in this request")
```
wardi: this would be spammy. I don't think it's necessary

ckanext/scheming_dynamic/schema_sync.py
```diff
@@ -0,0 +1,105 @@
...
+        return None
+
+    try:
+        # savepoint keeps a failed query from aborting the caller's transaction
+        with model.Session.begin_nested():
```
wardi: not necessary, we're only reading one value from the DB

ckanext/scheming_dynamic/schema_sync.py
```diff
@@ -0,0 +1,105 @@
...
+from sqlalchemy.exc import DBAPIError, UnboundExecutionError
+
+from ckan import model
+
+from ckanext.scheming_dynamic.model import SchemingSchema
+
+log = logging.getLogger(__name__)
+
+_fingerprint: tuple[int, datetime | None] | None = None
+_pending_fingerprint: tuple[int, datetime | None] | None = None
```
wardi: Let's store these as class variables in the _SchemingMixin class since they affect the data stored there, or as class variables in SchemingDynamicPlugin if you prefer

ckanext/scheming_dynamic/schema_sync.py
```diff
@@ -0,0 +1,105 @@
...
+    try:
+        # savepoint keeps a failed query from aborting the caller's transaction
+        with model.Session.begin_nested():
+            fingerprint = _current_fingerprint()
+    except (DBAPIError, UnboundExecutionError):
+        log.debug("cannot read the scheming_schema table")
+        return None
+
+    if fingerprint == _fingerprint:
+        log.debug("no database changes since last check")
```
wardi: also spammy

ckanext/scheming_dynamic/schema_sync.py
```diff
@@ -0,0 +1,105 @@
...
+
+def reset() -> None:
+    """Forget the cached fingerprint, forcing a re-merge on next access."""
+    global _fingerprint, _pending_fingerprint  # noqa: PLW0603
+    _fingerprint = None
+    _pending_fingerprint = None
+
+
+def _current_fingerprint() -> tuple[int, datetime | None]:
+    # count catches deleted rows, max(updated) catches created/changed ones
```
wardi: count doesn't catch deleted rows if we delete then create before the next check.

Instead, when we delete something we could insert a sentinel SchemingSchema that is empty except for an `updated` value.

ckanext/scheming_dynamic/validator.py
```diff
@@ -0,0 +1,32 @@
...
+
+import yaml
+from jsonschema import Draft202012Validator, FormatChecker
+from jsonschema.exceptions import ValidationError
+
+from ckanext.scheming_dynamic.schema import BaseSchema
+
+
+def load_data(path: Path) -> Any:
+    text = path.read_text()
```
wardi: Guard against systems with different default encodings
```suggestion
    text = path.read_text(encoding='utf-8'))
```

ckanext/scheming_dynamic/validator.py
```diff
@@ -0,0 +1,32 @@
...
+    if path.suffix in (".yaml", ".yml"):
+        return yaml.safe_load(text)
+
+    return json.loads(text)
+
+
+def iter_errors(data: Any, schema: BaseSchema) -> Iterator[ValidationError]:
+    built = schema.build()
+    Draft202012Validator.check_schema(built)
+    validator = Draft202012Validator(built, format_checker=FormatChecker())
```
wardi, line 29: does the validator support a way to provide translations of error messages so that errors on the front end are displayed in the user's language?

ckanext/scheming_dynamic/validator.py
```diff
@@ -0,0 +1,32 @@
...
+    yield from sorted(validator.iter_errors(data), key=lambda e: list(e.absolute_path))
+
+
+def error_location(error: ValidationError) -> str:
+    return "/".join(str(p) for p in error.absolute_path) or "<root>"
```
wardi:
```suggestion
    return "/".join(str(p) for p in error.absolute_path) or _("<root>")
```

setup.py
```diff
@@ -27,6 +27,7 @@
     install_requires=[
         'pyyaml',
         'ckanapi',
+        'jsonschema',
```
wardi: WDYT, should we make scheming_dynamic and this requirement an option? i.e.
```
pip install ckanext-scheming[dynamic]
```

wardi (commented):
Amazing work. Mostly minor comments with a few small changes requested.

ckanext/scheming/ckan_formpages_draft.yaml
mutantsan: This is unrelated, but it is a bug fix. Duplication of a `validators` property. I'd suggest keeping it.

ckanext/scheming_dynamic/logic/validators.py
mutantsan: As I can see, CKAN's `_validate` function sorts keys with the `flattened_order_key` function, so `entity_type` should already be there. No reason to double-check.

ckanext/scheming_dynamic/migration/scheming_dynamic/alembic.ini
mutantsan: Sure. Just a default auto-generated `alembic.ini`. Generated with the command:
```
ckan generate migration -p scheming_dynamic -m "Create scheming_dataset_schema table"
```

ckanext/scheming_dynamic/cli.py
mutantsan: You're right. It should be `entity_type`. But I suggest keeping group and organization validation support. Cause we do support the validation.

ckanext/scheming_dynamic/cli.py
mutantsan: Changed wording a bit to follow the ruff format.

ckanext/scheming_dynamic/cli.py
mutantsan: I'd suggest `validation_schema`, as we have only 1 of this kind. And having a long command isn't really convenient.

ckanext/scheming_dynamic/README.md
mutantsan: If I understood you right... JSON Schema validates the shape of the definition alone. It doesn't know about the `schema_type` or how it's going to be stored.

ckanext/scheming_dynamic/README.md
mutantsan: I've implemented the `start_form_page` support afterward and removed this line.

ckanext/scheming_dynamic/README.md
mutantsan: Yes. For now, I propose keeping it like that for the MVP. It's noted, so we can implement it later.

ckanext/scheming_dynamic/schema.py
mutantsan: I've seen non-string values in our schemas, so I assume it's better to support them.

image

ckanext/scheming_dynamic/schema.py
mutantsan: https://json-schema.org/draft/2020-12/json-schema-core#section-8.2.1

Empty fragments aren't allowed in the $id.

ckanext/scheming_dynamic/validator.py
mutantsan: No. We would have to build something on top of that, but I think that's beyond the scope of the current PR.

ckanext/scheming_dynamic/logic/validators.py
wardi: If the entity type passed is not supported we should fail with a validation error here.

ckanext/scheming_dynamic/README.md
mutantsan: I reworked the approach a bit. Now we have a simple separate table for tracking changes.

ckanext/scheming_dynamic/README.md
wardi: What I mean is we should have a ValidationError if the `schema_type` value passed doesn't match the `dataset_type`/`organization_type`/`group_type`.

One way of enforcing that a ValidationError is by raised when they don't match is by generating a JSON Schema that only allows one value for the type field and letting the JSON Schema validation report the error. Another way is to create a second validator.

wardi: An alternative is to drop the `schema_type` parameter and instead extract it from the `schema`. Then there's no error to handle.

ckanext/scheming_dynamic/schema.py
wardi: On forms the only thing we can put into choice values are strings. It doesn't make sense to allow numbers, bools or objects if they can't be passed by the form.

What's the use case?

ckanext/scheming_dynamic/schema.py
wardi: Without the fragment is good too.

ckanext/scheming_dynamic/logic/validators.py
mutantsan: I've changed the approach to validation a bit. Can you review it again?

ckanext/scheming_dynamic/README.md
mutantsan: It was validated in the `_validate_definition` function. But I think you're right that it's better to infer the `schema_type` from the `definition`. 

I also moved the definition validation into a proper CKAN validator.

ckanext/scheming_dynamic/schema.py
mutantsan: I feel a bit conflicted about this one. From what I've learned from the JSON Schema specification, `$id` isn't required to be a resolvable URL. It's primarily a unique identifier for the schema and serves a technical role (for example, as the canonical identifier and base URI for resolving relative references). It's not intended as user-facing documentation.

If we point it at the dataset schema documentation, what should we use for the group and organization schemas? It feels like we'd be using `$id` as a documentation link rather than a schema identifier, which doesn't seem to align with its intended purpose.

ckanext/scheming/plugins.py
```diff
@@ -72,6 +72,11 @@ class _SchemingMixin(object):
     _schemas = tuple()
     _expanded_schemas = tuple()
 
+    dynamic_scheming: dict[str, dict[str, Any]] = {
+        "schema": {"fingerprint": None, "pending_fingerprint": None},
+        "preset": {"fingerprint": None, "static": None},
+    }
```
wardi: it looks like the keys here don't change. For type checking and keeping the code simpler it would be better to drop the dicts and use 4 attributes, something like:

```python
    _dynamic_schema_fingerprint: str | None = None
    _dynamic_schema_pending_fingerprint: str | None = None
    _dynamic_preset_fingerprint: str | None = None
    _dynamic_preset_static: str | None = None
```

ckanext/scheming/plugins.py
```diff
@@ -237,21 +249,21 @@ def _sync_dynamic_schemas(self):
         if not p.plugin_loaded('scheming_dynamic'):
             return
 
-        from ckanext.scheming_dynamic import schema_sync # noqa
+        from ckanext.scheming_dynamic import sync # noqa
 
-        merged = schema_sync.dataset_schemas_if_changed(self._static_schemas)
+        merged = sync.dataset_schemas_if_changed(self._static_schemas)
```
wardi: won't this fail for normal scheming users if the dynamic plugin isn't enabled/migrations not applied?

ckanext/scheming/plugins.py
```diff
@@ -237,21 +249,21 @@ def _sync_dynamic_schemas(self):
...
         if merged is None:
             return
 
         try:
             expanded = _expand_schemas(merged)
-        except SchemingException:
+        except Exception:
```
wardi: We want unexpected exceptions to bubble up so they can be logged and fixed.

ckanext/scheming_dynamic/assets/js/scheming-schema-editor.js
```diff
@@ -1,23 +1,26 @@
...
- * scheming meta-schema (JSON Editor, loaded from a CDN).
+ * scheming meta-schema (JSON Editor).
  *
  * The textarea stays the source of truth for the POSTed value: the module
  * keeps it in sync and falls back to plain textarea editing when JSON
  * Editor is unavailable or the current content is not parseable JSON.
  */
 ckan.module('scheming-schema-editor', function ($) {
   return {
+    supportedLanguages: ['en', 'de', 'it', 'es'],
```
wardi, line 11: why are these hard-coded?

ckanext/scheming_dynamic/logic/action.py
```diff
@@ -1,48 +1,66 @@
 from __future__ import annotations
 
-from typing import Any
+from contextlib import nullcontext
+from typing import Any, cast
+
+from click import get_current_context
+from flask import has_app_context
```
wardi: It's very strange to need the click and flask context inside an action. What's this for?

ckanext/scheming_dynamic/logic/action.py
```diff
@@ -64,15 +82,60 @@ def scheming_schema_update(context: Any, data_dict: dict[str, Any]) -> dict[str,
...
-    _validate_definition(definition, entity_type, schema_type)
+    type_field = schema.TYPE_FIELDS[entity_type]
+    if definition[type_field] != schema_type:
+        raise tk.ValidationError(
+            {
+                "definition": [
+                    tk._(f"'{type_field}' must match schema_type '{schema_type}'")
+                ]
+            }
+        )
```
wardi: We could drop schema_type from update like we did for create so we don't need this check

ckanext/scheming_dynamic/logic/action.py
```diff
@@ -64,15 +82,60 @@ def scheming_schema_update(context: Any, data_dict: dict[str, Any]) -> dict[str,
...
+    -- so this locks ``definition`` as a new version instead. Otherwise
+    nothing depends on the head version yet, so it's safe to overwrite its
+    definition directly.
+
+    Returns the version row that now holds ``definition``.
+    """
+    head_version = SchemingSchemaVersion.head_version(entity_type, schema_type)
 
-    schema.update_definition(definition)
+    if SchemingSchemaPin.is_version_locked(entity_type, schema_type, head_version):
```
wardi: this will mostly work but is unsafe with parallel updates.

ckanext/scheming_dynamic/logic/action.py
```diff
@@ -89,30 +152,180 @@ def scheming_schema_delete(context: Any, data_dict: dict[str, Any]) -> bool:
...
+    if not head:
         raise tk.ObjectNotFound(tk._(f"Schema for '{schema_type}' not found"))
 
-    schema.delete()
+    SchemingSchemaActivity.record(
+        entity_type,
+        schema_type,
+        SchemingSchemaActivity.DELETE,
+        context["user"],
+        head.definition,
```
wardi, line 191: If every extension creates its own models that are like the activity feed ones then it's much harder to centralize a retention policy and manage database bloat

ckanext/scheming_dynamic/logic/action.py
```diff
@@ -89,30 +152,180 @@ def scheming_schema_delete(context: Any, data_dict: dict[str, Any]) -> bool:
...
+    preset = SchemingPreset.get(preset_name)
+    if not preset:
+        raise tk.ObjectNotFound(tk._(f"Preset '{preset_name}' not found"))
+
+    if definition["preset_name"] != preset_name:
         raise tk.ValidationError(
-            {"definition": [tk._(f"'{type_field}' must match schema_type '{schema_type}'")]}
+            {
+                "definition": [
+                    tk._(f"'preset_name' must match preset_name '{preset_name}'")
```
wardi: Instead of passing the preset_name and the values in the definition let's just pass the values. Presets aren't free-form like dataset schemas with arbitrary fields defined at the top level.

ckanext/scheming_dynamic/logic/schema.py
```diff
@@ -5,23 +5,35 @@
...
     unicode_safe: types.Validator,
     convert_to_json_if_string: types.Validator,
+    scheming_default_entity_type: types.DataValidator,
+    scheming_definition_valid: types.DataValidator,
     one_of: types.ValidatorFactory,
     default: types.ValidatorFactory,
 ) -> types.Schema:
     return {
-        "schema_type": [not_missing, unicode_safe],
+        "__before": [scheming_default_entity_type],
```
wardi, line 18: this looks like it does the same thing as `default(DEFAULT_ENTITY_TYPE)` below

ckanext/scheming_dynamic/templates/package/new_resource_not_draft.html
```diff
@@ -0,0 +1,5 @@
+{% ckan_extends %}
+
+{% block form %}
+  {% snippet resource_form_snippet, data=data, errors=errors, error_summary=error_summary, include_metadata=false, pkg_name=pkg_name, pkg_id=pkg.id, stage=stage, dataset_type=dataset_type %}
+{% endblock %}
```
wardi, line 5: These three templates are extended just to pass pkg_id through?

ckanext/scheming_dynamic/views/admin.py
```diff
@@ -113,16 +134,147 @@ def post(self, schema_type: str) -> str | Any:
...
+            css_class = "diff-meta"
+        elif line.startswith("@@"):
+            css_class = "diff-hunk"
+        elif line.startswith("+"):
+            css_class = "diff-add"
+        elif line.startswith("-"):
+            css_class = "diff-del"
+        else:
+            css_class = "diff-ctx"
+        lines.append(Markup('<span class="{}">{}</span>').format(css_class, line))
```
wardi, line 302: line needs to be escaped here

ckanext/scheming_dynamic/model.py
```diff
@@ -1,72 +1,448 @@
...
+class SchemingSchemaVersion(tk.BaseModel):
+    """Schema version history, plus the live definition."""
+
+    __table__ = sa.Table(
+        "scheming_schema_version",
+        tk.BaseModel.metadata,
+        sa.Column("entity_type", sa.Text, primary_key=True),
+        sa.Column("schema_type", sa.Text, primary_key=True),
+        sa.Column("version", sa.Integer, primary_key=True),
+        sa.Column("definition", JSONB, nullable=False),
```
wardi, line 162: do we also need an expanded_definition that includes the expanded presets at the time that this version was created?

ckanext/scheming_dynamic/model.py
```diff
@@ -1,72 +1,448 @@
...
+            .all()
+        )
+
+    @classmethod
+    def lock(
         cls, entity_type: str, schema_type: str, definition: dict[str, Any]
-    ) -> SchemingSchema:
+    ) -> SchemingSchemaVersion:
+        """Snapshot ``definition`` as the next version for this schema_type."""
+        version = cls.head_version(entity_type, schema_type) + 1
```
wardi: race condition and unnecessary work here. We could use a `server_default` on the column instead to avoid this.

ckanext/scheming_dynamic/preset_resolve.py
```diff
@@ -0,0 +1,57 @@
+from __future__ import annotations
+
+from typing import Any
+
+
+class PresetCycleError(Exception):
+    """A preset's ``preset`` (base) chain loops back on itself."""
```
wardi, line 7: as suggested in -26 let's not use "preset" inside preset values, an "extends" outside the values would be cleaner.

ckanext/scheming_dynamic/README.md
```diff
@@ -36,44 +56,73 @@ Apply the migration:
...
+### Enabling on a portal with existing data
+
+If the portal already has datasets under a file-defined (`scheming.dataset_schemas`)
+type, importing that schema into the database and pinning existing datasets to it
+keeps them validating/rendering against the exact schema they were created with,
+even after later edits. Run once per existing dataset type:
+
+```sh
+ckan scheming-dynamic sync --type dataset SCHEMA_TYPE
+ckan scheming-dynamic pin --type dataset SCHEMA_TYPE
```
wardi, line 34: would it be better to have one command for both?

ckanext/scheming/plugins.py
mutantsan: That's why we have log.exception, which will print out the traceback

ckanext/scheming_dynamic/assets/js/scheming-schema-editor.js
mutantsan: Cause these are what supported by the library. It's possible to add support of other languages, which is our of scope https://germanbisurgi.github.io/jedison-docs/language-and-translations

ckanext/scheming_dynamic/logic/action.py
mutantsan: Your point?

ckanext/scheming_dynamic/templates/package/new_resource_not_draft.html
mutantsan: It seems that this is the only way. If the ID or the whole package had been passed initially, I wouldn't have to do it. Name isn't a reliable identifier.

ckanext/scheming_dynamic/logic/action.py
mutantsan: I'm not afraid of an extra table. This is what a database is for. Using the activity plugin for it means it can be... plugged off. Then we lose a bunch of features - revert and history. It also can be cleared from elsewhere. We also will have to fight its permission system.

ckanext/scheming/plugins.py
wardi: *unexpected* exceptions should cause a server error, not add a log message and continue. If we're suppressing exceptions and logging them it shouldn't be for all exception types

ckanext/scheming_dynamic/assets/js/scheming-schema-editor.js
wardi: ok,please add a comment indicating where these language choices come from (i.e. the library we're using)

ckanext/scheming_dynamic/model.py
wardi: I think this is a simpler and less confusing option than versioning presets

ckanext/scheming_dynamic/logic/action.py
wardi: I'm not afraid of tables, I'm trying to avoid duplicated functionality that needs to be maintained, documented and managed separate from core functionality

ckanext/scheming_dynamic/templates/package/new_resource_not_draft.html
wardi: I'm +1 on updating the core templates to always pass through the whole package to these snippets. ckan 3.0 is an opportunity to make the templates and snippets more consistent

README.md
```diff
@@ -570,6 +570,39 @@ markdown text area and display
 You may define your own presets by adding additional files to the `scheming.presets`
 [configuration setting](#configuration).
 
+#### Restricting where a preset may be used
+
+Some presets only make sense on one particular field, or depend on other
+keys being present on the field. A preset may declare these constraints in
+its `values`. They are checked when schemas are expanded at startup, and a
```
wardi: These should be one level up (as siblings of `preset_name` and `values`) , not mixed with values. values are the things actually applied to fields when using a preset, but `restrict_to_field` and `requires_one_of` are restrictions on the fields that presets may be applied to.

README.md
```diff
@@ -570,6 +570,39 @@ markdown text area and display
...
+The preset may only be applied to the field named `field_name` on a schema
+for the given `entity_type` (`dataset`, `resource`, `group` or
+`organization`). Applying it to any other field raises an error. This is
+used by the core presets that wrap CKAN's built-in fields, for example
+`dataset_slug` (dataset `name`), `dataset_organization` (dataset
+`owner_org`), `resource_url_upload` (resource `url`) and
+`organization_url_upload` (organization `image_url`).
+
+```yaml
+  requires_one_of: [choices, choices_helper]
```
wardi: For this to be generic and support requirig fields other than choices, we need something more structured.  How about:
```yaml
requires:
 - [choices, choices_helper]
 - other_field
```
to mean "requires one of choices, choices_helper, and also requires other_field"

README.md
```diff
@@ -570,6 +570,39 @@ markdown text area and display
...
+#### Restricting where a preset may be used
+
+Some presets only make sense on one particular field, or depend on other
+keys being present on the field. A preset may declare these constraints in
+its `values`. They are checked when schemas are expanded at startup, and a
+violation raises a `SchemingException` (so the schema fails fast rather than
+rendering a broken form).
+
+```yaml
+  restrict_to_field: {entity_type: dataset, field_name: title}
```
wardi, line 593: entity type should support a list of values

ckanext/scheming_dynamic/README.md
```diff
@@ -0,0 +1,221 @@
...
+the run continues.
+
+Only one run may be active per version pair at a time, enforced by a partial
+unique index. Applying a mapping is idempotent: a dataset already on the target
+version is recorded `skipped`, so a run killed halfway is resumed by running it
+again.
+
+### Data loss
+
+There is no revert. Datasets are meant to converge on the live schema, and the
```
wardi: if activities are enabled a copy of the metadata is available that could be used for reverting migrations

ckanext/scheming_dynamic/schema_migration/apply.py
```diff
@@ -0,0 +1,294 @@
...
+        return types.Context(
+            user=self.user,
+            ignore_auth=True,
+            session=model.Session,
+        )
+
+    def _read_context(self) -> types.Context:
+        """Read the dataset the way the edit form does.
+
+        Without ``for_edit`` an uploaded resource's ``url`` comes back fully
```
wardi, line 251: This reminds me that we should fix this upstream and either make this a normal parameter or make it the default in 3.0. @JVickery-TBS had a similar issue with this and dynamic hostnames.

README.md
mutantsan: \+

ckanext/scheming_dynamic/README.md
mutantsan: Don't think it's necessary. Activity is also a separate plugin.

ckanext/scheming_dynamic/schema_migration/apply.py
mutantsan: Not sure what you meant here. I'll leave a `TODO` in the code so we can revisit it if the core behavior changes.

README.md
mutantsan: \+

README.md
mutantsan: \+

ckanext/scheming_dynamic/README.md
mutantsan: I don't think so. The `sync` is metadata-only: locks a new schema version, fast, idempotent, safe to run on every deploy.

The `pin` walks every entity of that type, validates each, and writes a pin row. It's a one-time bulk migration you'd run behind `--dry-run` first.

They also have non overlapping flags.

ckanext/scheming_dynamic/model.py
mutantsan: \+

ckanext/scheming_dynamic/views/admin.py
mutantsan: It seems like the `Markup(...).format` already runs escape

image

ckanext/scheming_dynamic/logic/schema.py
mutantsan: The `scheming_definition_valid` needs `entity_type` to know which schema to validate against. `navl` validates keys in sorted (alphabetical) order, not declaration order, so definition is checked before `entity_type`. That means taht the per-key `default(DEFAULT_ENTITY_TYPE)` hasn't run yet. The `__before` sets the default for the whole dict up front, so `entity_type` is always resolved before definition is validated.

ckanext/scheming_dynamic/logic/action.py
mutantsan: \+

ckanext/scheming_dynamic/logic/action.py
mutantsan: As I said, `activity` is a separate plugin. It's not a part of the CKAN core and it can be disabled, then... `we lose a bunch of features - revert and history. It also can be cleared from elsewhere. We also will have to fight its permission system.`

ckanext/scheming_dynamic/logic/action.py
mutantsan: \+

ckanext/scheming_dynamic/logic/action.py
mutantsan: \+

